Issue #41, opened 2026-02-16
fetch_markets() - how to ensure we are not drifting/reordering between calls?


lundgrenolof:
When using the "offset" option, how do we know markets haven't drifted/reordered between calls? There doesnt seem to be a way to account for that unless we have a cursor where we paginate results properly?

PR #45, opened 2026-02-18
feat: Add cursor-based pagination to fetchMarkets

CrispyyBaconx:
addresses #41

- Added cursor-based pagination support in BaseExchange via a new fetchMarketsPaginated method, using snapshot-backed cursors for stable pagination across calls.
- Introduced a reusable PaginatedResult<T> type to standardize paginated responses (data, nextCursor, total).
Kept fetchMarkets backward-compatible (still returns UnifiedMarket[]) while enabling cursor-based pagination through the new paginated method.
- Updated exchange integrations and constructor wiring (Baozi, Kalshi, Limitless, Polymarket, plus shared base behavior) to support configurable pagination snapshot TTL and unified options handling with legacy constructor compatibility.
- Updated OpenAPI documentation to preserve the existing fetchMarkets response shape and add the new fetchMarketsPaginated endpoint.
- Added/updated unit and compliance tests to validate cursor pagination behavior (consistency, cursor expiry, fallback behavior) and confirm backward compatibility.

### Notes

- This would change the api surface a bit for fetchMarkets since the paginated version returns the new paginated object, using a separate fetchMarketsPaginated for now as to not be a breaking change. Maybe in v3.0.0 can be replaced
- I don't know whats better ergonomically wrt the PredictionMarketExchange constructor, whether it should have 2 params for the creds + config options or if it should just be 1 object where things could be added in the future? 
- There is a tradeoff in data freshness + the stability of being able to navigate the list, maybe can add some way to force refresh every time or a separate method to always get fresh data for people that want new markets immediately?

### DIFF
--- a/core/src/BaseExchange.ts
+++ b/core/src/BaseExchange.ts
@@ -1,4 +1,4 @@
-import { UnifiedMarket, UnifiedEvent, PriceCandle, CandleInterval, OrderBook, Trade, Order, Position, Balance, CreateOrderParams } from './types';
+import { UnifiedMarket, UnifiedEvent, PriceCandle, CandleInterval, OrderBook, Trade, Order, Position, Balance, CreateOrderParams, PaginatedResult } from './types';
 import { getExecutionPrice, getExecutionPriceDetailed, ExecutionPriceResult } from './utils/math';
 import { MarketNotFound, EventNotFound } from './errors';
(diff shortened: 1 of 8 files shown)


realfishsam:
I'm determining the best way to merge this 🤔 .

What's your Discord handle, btw?

Issue #90, opened 2026-04-14
fix: AddressWatcher empty catch block silently swallows all subscription errors


realfishsam:
## Bug

`src/subscriber/watcher.ts`, lines 204-205:

```typescript
try {
    // ~30 lines: buildActivity, fetchFn, getChanged, resolver dispatch
} catch {
}
```

A completely empty catch block wraps the entire \`handleSubscriptionData\` pipeline. Every possible failure — network errors, parse errors, type mismatches, data validation — is silently discarded.

### Promises hang forever
When an error is swallowed, pending \`watch()\` callers are **never resolved or rejected**. They remain in the \`resolvers\` map indefinitely:

```typescript
const snapshot = await exchange.watchAddress(address, types); // HANGS FOREVER
```

### State corruption
\`lastState\` may be partially updated. Subsequent events compare against corrupted state via \`getChanged()\`, delivering wrong data.

### Resource leak
Resolver promises accumulate without cleanup. Memory pressure increases over time.

### Double-catch chain
This is the **second layer** of error suppression. Issues in Polymarket/Limitless \`fetchWatchedAddressActivity\` (silent \`.catch()\` handlers) swallow errors at the exchange layer. Even if those are fixed, this empty catch would still suppress any errors that propagate. **Both layers must be fixed together.**

Ironically, the GoldSky layer below (goldsky.ts:348-371) DOES handle errors with logging. The watcher above it discards everything.

## Fix

```typescript
} catch (err) {
    const error = err instanceof Error ? err : new Error(String(err));
    console.error(\`[AddressWatcher] Error for \${address}:\`, error);

    const resolvers = this.resolvers.get(key);
    if (resolvers?.length) {
        resolvers.forEach(r => r.reject(error));
        this.resolvers.set(key, []);
    }
    // Also reject asset-specific resolvers
}
```

## Related

- Must be fixed with Polymarket silent catch issue and Limitless silent catch issue
- Part of fallback audit: `core/docs/fallback-audit/07-watcher-empty-catch.md`

realfishsam:
Fixed in e8c0010 — empty catch replaced with proper error rejection

PR #94, opened 2026-04-15
fix spurious nonetwork on watch_address balance rpc

### LINKED ISSUE #92: Consistent noNetwork errors when calling watch_address on 2.30.7
I can't repro on 2.30.6 but can consistently repro on 2.30.7, so I suspect the culprit is the fix for #90.

Here's the error:

```
pmxt.errors.PmxtError: could not detect network (event="noNetwork", code=NETWORK_ERROR, version=providers/5.8.0)
```

I don't have a full stack trace because the exception is raised in an async task, but I can get a stack trace if it would be helpful. That said, I'm not sure how helpful a stack trace would be since it would probably hit a singularity at the boundary between my Python process and the server process.

Also, for what it's worth, my machine (and Python process) is connected to the network, so I think the error itself is spurious.

mooncitydev:
Here’s a cleaner, sharper version:

---

## What changed

In ethers v5, `JsonRpcProvider` performs automatic network detection (`eth_chainId`) on first use. Public RPC endpoints can fail or rate-limit this call, causing `noNetwork` / `NETWORK_ERROR` even when connectivity is fine.

After #90, these errors surfaced because they’re no longer swallowed in `fetchWatchedAddressActivity`.

To fix this, we now pass a static network (chainId + name) for Polygon and Base, bypassing auto-detection and using the RPC strictly for contract reads.

## Files

* `core/src/exchanges/polymarket/index.ts` — Polygon (137)
* `core/src/exchanges/limitless/index.ts` — Base (8453)
* `core/src/exchanges/limitless/client.ts` — apply same static network for `getBalance()`

Closes #92

---

**Summary:** Use static network config to avoid RPC `eth_chainId` failures and eliminate intermittent `NETWORK_ERROR`.

### DIFF
--- a/core/src/exchanges/limitless/client.ts
+++ b/core/src/exchanges/limitless/client.ts
@@ -176,7 +176,10 @@ export class LimitlessClient {
         const ABI = ["function balanceOf(address) view returns (uint256)", "function decimals() view returns (uint8)"];
         
         // Use a public RPC for Base
-        const provider = new providers.JsonRpcProvider('https://mainnet.base.org');
+        const provider = new providers.JsonRpcProvider('https://mainnet.base.org', {
+            chainId: 8453,
+            name: 'base',
+        });
         const contract = new Contract(USDC_ADDRESS, ABI, provider);
 
         const balance = await contract.balanceOf(this.signer.address);
--- a/core/src/exchanges/limitless/index.ts
+++ b/core/src/exchanges/limitless/index.ts
@@ -463,7 +463,12 @@ export class LimitlessExchange extends PredictionMarketExchange {
 
     private async getAddressOnChainBalance(targetAddress: string): Promise<Balance[]> {
         // Query USDC balance directly from Base chain
-        const provider = new providers.JsonRpcProvider('https://mainnet.base.org');
+        // Static network avoids ethers v5 auto-detect (eth_chainId) which can throw
+        // noNetwork / NETWORK_ERROR on flaky public RPCs (#92).
+        const provider = new providers.JsonRpcProvider('https://mainnet.base.org', {
+            chainId: 8453,
+            name: 'base',
+        });
 
         // Get USDC contract address for Base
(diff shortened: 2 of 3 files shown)


realfishsam:
Thanks for the PR — the diagnosis is correct. The `eth_chainId` auto-detection on `JsonRpcProvider` is exactly what's causing the spurious `noNetwork` errors after #90 surfaced them.

However, the fix as written doesn't actually work. In ethers v5 (which we're on — `ethers/5.8.0`), passing a network object to `JsonRpcProvider` stores it internally but **still calls `detectNetwork()` → `eth_chainId`** on first use. We verified this locally by intercepting RPC calls — both with and without the static network argument, `JsonRpcProvider` makes the same `eth_chainId` call and fails the same way when the public RPC drops it.

The ethers v5 mechanism for skipping auto-detection is `StaticJsonRpcProvider`:

```typescript
// This still calls eth_chainId — does NOT fix the issue
const provider = new ethers.providers.JsonRpcProvider(url, { chainId: 137, name: 'matic' });

// This skips eth_chainId entirely — the actual fix
const provider = new ethers.providers.StaticJsonRpcProvider(url, { chainId: 137, name: 'matic' });
```

We confirmed `StaticJsonRpcProvider` goes straight to `eth_call` with zero detection calls, even when the RPC endpoint is flaky.

To get this merged, the change needed is straightforward — swap `JsonRpcProvider` → `StaticJsonRpcProvider` in all three locations:
- `core/src/exchanges/polymarket/index.ts` (Polygon, chainId 137)
- `core/src/exchanges/limitless/index.ts` (Base, chainId 8453)
- `core/src/exchanges/limitless/client.ts` (Base, chainId 8453)

Everything else in the PR (the network objects, the comments, the scope) is good. Happy to re-review once updated.

ndmeiri:
Thanks for taking a look at this @mooncitydev. To unblock the fix, I opened #100 and added you as an author.

PR #100, opened 2026-04-21
Fix spurious balance fetch errors

### LINKED ISSUE #92: Consistent noNetwork errors when calling watch_address on 2.30.7

ndmeiri:
Updates relevant exchange adapters to use `StaticJsonRpcProvider`, which is the ethers v5 mechanism to skip automatic network detection.

Note that when PMXT eventually upgrades to ethers v6, we'll be able to migrate back to `JsonRpcProvider`. (See the [migration guide](https://github.com/ethers-io/ethers.js/blob/b746c3cf6cd191e2357fa55751696676ffda060e/docs.wrm/migrating.wrm#L231-L244) and https://github.com/ethers-io/ethers.js/discussions/3994#discussioncomment-5710699).

Adapted from @mooncitydev's #94.

Closes #92.

### DIFF
--- a/core/src/exchanges/limitless/client.ts
+++ b/core/src/exchanges/limitless/client.ts
@@ -176,7 +176,10 @@ export class LimitlessClient {
         const ABI = ["function balanceOf(address) view returns (uint256)", "function decimals() view returns (uint8)"];
         
         // Use a public RPC for Base
-        const provider = new providers.JsonRpcProvider('https://mainnet.base.org');
+        const provider = new providers.StaticJsonRpcProvider('https://mainnet.base.org', {
+            chainId: 8453,
+            name: 'base',
+        });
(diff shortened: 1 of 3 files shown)


realfishsam:
Good work

PR #108, opened 2026-04-23
fix opinion child markets hardcoding volume24h to 0

### LINKED ISSUE #106: Opinion: child markets hardcode volume24h to 0
## Problem

The Opinion normalizer hardcodes `volume24h: 0` for child markets (outcomes within categorical/multi-outcome events), even when the parent market is actively traded.

`core/src/exchanges/opinion/normalizer.ts:303`:
```typescript
volume24h: 0, // child markets do not have volume24h
```

The Opinion API only returns `volume24h` on parent markets, not on children. But since children inherit the parent's trading activity, they shouldn't be marked as zero.

## Impact

Downstream consumers (e.g. matching pipelines) that filter on `volume24h > 0` skip these markets entirely. This contributes to Opinion's near-zero embedding coverage in the matching engine (0.4%).

## Suggested fix

Distribute the parent's `volume24h` across child markets, either evenly or proportionally by price/volume. Alternatively, copy the parent's `volume24h` to each child so they aren't filtered out.

mooncitydev:
## what's the problem

child markets (outcomes within categorical/multi-outcome events) always had `volume24h` set to `0` in the Opinion normalizer, even when the parent market was actively traded. this caused downstream consumers that filter on `volume24h > 0` (like matching pipelines) to skip these markets entirely â€” which is why Opinion had near-zero embedding coverage (~0.4%).

## what changed

instead of hardcoding `volume24h: 0` for child markets, we now read the parent's `volume24h` and distribute it evenly across all children. this is a pragmatic approximation since the Opinion API only provides volume at the parent level.

`normalizeMarketsFromEvent` computes `parentVolume24h / childCount` and passes that into each `normalizeChildMarket` call. the private method now accepts a `volume24h` parameter (defaults to `0` to stay safe for any direct callers).

## why even distribution?

the Opinion API doesn't expose per-child volume, so we can't be more precise here. splitting evenly is honest and avoids artificially inflating or hiding any specific outcome. it also unblocks the matching pipeline immediately without needing a more complex weighting scheme.

fixes #106

made by mooncitydev

### DIFF
--- a/core/src/exchanges/opinion/normalizer.ts
+++ b/core/src/exchanges/opinion/normalizer.ts
@@ -57,8 +57,11 @@ export class OpinionNormalizer implements IExchangeNormalizer<OpinionRawMarket, 
         const children = raw.childMarkets || [];
         const results: UnifiedMarket[] = [];
 
+        const parentVolume24h = parseNumStr(raw.volume24h);
+        const perChildVolume24h = children.length > 0 ? parentVolume24h / children.length : 0;
+
         for (const child of children) {
-            const market = this.normalizeChildMarket(child, raw);
+            const market = this.normalizeChildMarket(child, raw, perChildVolume24h);
             if (market) results.push(market);
         }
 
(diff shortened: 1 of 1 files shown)


realfishsam:
Merged to main in e5478643fc66c35fd57d09d9e35ad278136ce67e — changed the distribution strategy from even split to proportional by child volume. Thanks for the contribution!

PR #116, opened 2026-04-27
fix(polymarket): honor OpenAPI servers override for geoblock

### LINKED ISSUE #115: Polymarket geoblock endpoint
Is there any interest in adding support for [Polymarket's geoblock endpoint](https://docs.polymarket.com/api-reference/geoblock#geoblock-endpoint)?

If so, in which spec file in `core/specs/polymarket/` should it live? (Or perhaps a new spec file because the base URL, `https://polymarket.com/api`, differs from that of the Data, Gamma, and CLOB APIs.)

It's a simple (i.e., zero-argument) GET request that clients could handle on their own, but it would be nice for PMXT to abstract away the call since it's already handling nearly every other Polymarket API call.

mooncitydev:
the geoblock endpoint lives at https://polymarket.com/api/geoblock while the CLOB spec defaults to clob.polymarket.com, which returns 404 for this path. OpenAPI supports path- and operation-level servers; parseOpenApiSpec now records those overrides on ApiEndpoint.baseUrl, and implicit API requests use them when resolving URLs.

- Set operationId getGeoblock for stable method naming
- Add unit tests for server override resolution

Fixes #115
moon

### DIFF
--- a/core/specs/polymarket/PolymarketClobAPI.yaml
+++ b/core/specs/polymarket/PolymarketClobAPI.yaml
@@ -954,6 +954,7 @@ paths:
   # ----------------------------------------------------------------------------
   /geoblock:
     get:
+      operationId: getGeoblock
       summary: Check Geoblock Status
       description: Check the geographic eligibility of the requesting IP address.
       tags: [System]
--- a/core/src/BaseExchange.ts
+++ b/core/src/BaseExchange.ts
@@ -31,6 +31,11 @@ export interface ApiEndpoint {
     isPrivate?: boolean;
     /** Identifier used to generate the implicit API method name. */
     operationId?: string;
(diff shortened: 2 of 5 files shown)


ndmeiri:
@realfishsam I know you've been super busy with v2 release. Any chance we could move toward getting this reviewed?

realfishsam:
> @realfishsam I know you've been super busy with v2 release. Any chance we could move toward getting this reviewed?

Yes! Sorry for the delay. It'll be reviewed tomorrow morning.

realfishsam:
> @realfishsam I know you've been super busy with v2 release. Any chance we could move toward getting this reviewed?

Reviewed and merged. Will be available in a new version within 10 minutes.

PR #511, opened 2026-05-24
fix: export Python websocket return types

### LINKED ISSUE #461: `SubscribedAddressSnapshot` and `FirehoseEvent` not exported from Python `__init__.py`
## Drift
`SubscribedAddressSnapshot` and `FirehoseEvent` are public return types used by `watch_address()` and `watch_all_order_books()` / `firehose()` respectively. Both are exported as public types from the TypeScript SDK via `export type * from "./pmxt/models.js"`. Neither is imported nor listed in `__all__` in the Python package's `__init__.py`, making them inaccessible at the `pmxt.*` namespace level.

## TypeScript SDK
`sdks/typescript/pmxt/models.ts`, lines 159–168 (`FirehoseEvent`) and lines 773–794 (`SubscribedAddressSnapshot`):
```typescript
export interface FirehoseEvent { ... }
export interface SubscribedAddressSnapshot { ... }
```
Re-exported from `sdks/typescript/index.ts` line 32: `export type * from "./pmxt/models.js";`

## Python SDK
`sdks/python/pmxt/models.py`, lines 278–289 (`FirehoseEvent`) and lines 469–487 (`SubscribedAddressSnapshot`) — both exist in `models.py`.

`sdks/python/pmxt/__init__.py`, lines 39–64 (import block): neither `FirehoseEvent` nor `SubscribedAddressSnapshot` is imported.
`sdks/python/pmxt/__init__.py`, lines 134–190 (`__all__` list): neither name appears.

## Expected
Both types should be imported in `__init__.py` and added to `__all__`:
```python
from .models import (
    ...
    FirehoseEvent,
    SubscribedAddressSnapshot,
)
__all__ = [
    ...
    "FirehoseEvent",
    "SubscribedAddressSnapshot",
]
```

## Impact
Python users who type-hint the return value of `watch_address()` or `watch_all_order_books()` / `firehose()` cannot write `pmxt.SubscribedAddressSnapshot` or `pmxt.FirehoseEvent` — they must reach into the internal `pmxt.models` module. IDEs and type checkers report these types as unavailable from the public package namespace, causing friction and potential breakage if the internal module path ever changes.

---

iccccccccccccc:
Fixes #461

## Changes
- export `FirehoseEvent` and `SubscribedAddressSnapshot` from the Python package entry point
- add a focused regression test for the Python public export list
- include the generator drift fixes from #509 so current `main` passes the exchange/generator checks

## Verification
- `python -m pytest sdks/python/tests/test_public_exports.py`
- `python -m py_compile sdks/python/pmxt/__init__.py sdks/python/pmxt/models.py sdks/python/tests/test_public_exports.py`
- `git diff --check`


realfishsam:

### What This Does
Fixes #461 by exporting `FirehoseEvent` and `SubscribedAddressSnapshot` from the Python SDK's `__init__.py` (both the import block and `__all__`). Additionally bundles two generator drift fixes from #509: (1) renames the Python class `Polymarket_us` to `PolymarketUS` to match the TypeScript SDK convention, and (2) adds a `fetchOrderBook` special case to both SDK generators so regeneration preserves the hand-written array-return handling.

### Blast Radius
- **Python SDK**: `__init__.py` (exports), `_exchanges.py` (class rename), generator script
- **TypeScript SDK**: generator script only (no runtime change to `client.ts`)
- **Core**: `generate-python-exchanges.js` (toClassName logic + CRLF regex)
- **Docs**: Both `API_REFERENCE.md` files regenerated (generated output, not hand-edits)

### Test Results
- Build: NOT RUN (bash permissions restricted in review environment)
- Unit tests: NOT RUN

### Findings

1. **`Polymarket_us` -> `PolymarketUS` is a breaking change for Python users** (`sdks/python/pmxt/_exchanges.py:368`, `sdks/python/pmxt/__init__.py:20,146`). Any Python user writing `pmxt.Polymarket_us()` will get an `ImportError` after upgrading. The TypeScript SDK already uses `PolymarketUS`, so this aligns the two SDKs, but there is no deprecation alias (e.g., `Polymarket_us = PolymarketUS`) to ease migration. Recommend adding a deprecation alias or documenting the rename in the changelog.

2. **fetchOrderBook generator special case duplicates existing hand-written code** (`sdks/python/scripts/generate-client-methods.js:390-424`, `sdks/typescript/scripts/generate-client-methods.js:384-412`). The current `client.py` and `client.ts` on `main` already contain the correct hand-written `fetchOrderBook` with array-return handling. The PR adds the same logic as a hardcoded special case in both generators so that running regeneration does not clobber it. This is correct and necessary -- the BaseExchange.ts declares the return type as `Promise<OrderBook>` (single), but the runtime can return `OrderBook[]` when `since+until` are provided, so the generic generator would produce broken code. However, the right long-term fix would be to update the BaseExchange.ts return type to `Promise<OrderBook | OrderBook[]>` and teach the generators about union return types.

3. **`toClassName` change has a special case only for `'us'`** (`core/scripts/generate-python-exchanges.js:33`). The new logic `part.toLowerCase() === 'us' ? 'US'` hardcodes a single abbreviation. If future exchanges have other abbreviations (e.g., `uk`, `eu`), each would need another special case. This is fine for now but worth noting.

4. **CRLF regex fix is defensive and correct** (`core/scripts/generate-python-exchanges.js:247`). The `\r?\n` change makes the `__all__` regex work on Windows line endings. No issue.

5. **Test `test_public_exports.py` is well-designed** (`sdks/python/tests/test_public_exports.py`). Uses AST parsing to verify imports and `__all__` entries without importing the package (which would require the sidecar). Good regression test.

6. **API_REFERENCE.md changes are consistent with regenerated output** -- the `testDummyMethod` -> `close` rename reflects that `testDummyMethod` was removed from `BaseExchange.ts` and `close()` was added. The `FetchOrderBookParams` type documentation was added. These are generated files.

### PMXT Pipeline Check
- Field propagation (3-layer): OK -- `FirehoseEvent` and `SubscribedAddressSnapshot` were already in `models.py` and `client.py` imports; only `__init__.py` was missing them
- OpenAPI sync: N/A -- no OpenAPI schema changes
- Type safety: OK -- no `any` types introduced, no non-null assertions

### Semver Impact
**minor** -- The `Polymarket_us` -> `PolymarketUS` rename is a breaking change for Python users, but `polymarket_us` is a relatively new exchange and the rename aligns with the established TypeScript convention. If any users depend on the old name, a deprecation alias should be added (which would make this a patch).

### Risk
- Tests were not run due to environment restrictions. The PR's verification steps (listed in the PR body) should be run by CI.
- The `Polymarket_us` -> `PolymarketUS` rename has no backwards-compatibility alias. Users upgrading will get a hard `ImportError` with no guidance.

iccccccccccccc:
Good catch on the Python class rename. I added a backwards-compatible Polymarket_us = PolymarketUS alias, kept it exported from pmxt.__all__, and updated the Python exchange generator so a future regen preserves the alias instead of dropping it. I also extended 	est_public_exports.py to cover that alias and reran python -m pytest sdks/python/tests/test_public_exports.py.

PR #566, opened 2026-05-24
fix: use Python 3.8-compatible annotations in errors.py

### LINKED ISSUE #561: [python-types] errors.py: Python 3.10+/3.9+ annotation syntax breaks Python 3.8 import
## File
`sdks/python/pmxt/errors.py`

## Problem

`errors.py` does **not** have `from __future__ import annotations`, which means all annotations are evaluated at import time. The file uses two syntax forms that are unavailable on Python 3.8:

### Python 3.10+ union syntax (`X | Y`) — Lines 12, 66, 84

| Line | Code | Fails on |
|------|------|----------|
| 12 | `exchange: str \| None = None` in `PmxtError.__init__` | Python 3.8, 3.9 |
| 66 | `retry_after: int \| None = None` in `RateLimitExceeded.__init__` | Python 3.8, 3.9 |
| 84 | `field: str \| None = None` in `ValidationError.__init__` | Python 3.8, 3.9 |

### Python 3.9+ built-in generic (`dict[K, V]`) — Line 102

| Line | Code | Fails on |
|------|------|----------|
| 102 | `_ERROR_CODE_MAP: dict[str, type[PmxtError]] = {...}` | Python 3.8 |

On Python 3.8, any `import pmxt` that transitively imports `errors.py` will raise:
```
TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
```
or
```
TypeError: 'type' object is not subscriptable
```

## Impact

- **CRITICAL**: The entire SDK fails to import on Python 3.8, despite `pyproject.toml` declaring `python_requires = ">=3.8"`.
- Every exchange class (`Polymarket`, `Kalshi`, etc.) is broken on Python 3.8 because `errors.py` is imported at startup.

## Suggested Fix

Add `from __future__ import annotations` at the top of the file (defers all annotation evaluation to string form, compatible with 3.8+) **and** replace the module-level `_ERROR_CODE_MAP` annotation with a typing-compatible form:

```python
from __future__ import annotations
from typing import Dict, Optional, Type

# Line 12
def __init__(self, ..., exchange: Optional[str] = None): ...

# Line 66
def __init__(self, ..., retry_after: Optional[int] = None, **kwargs): ...

# Line 84
def __init__(self, ..., field: Optional[str] = None, **kwargs): ...

# Line 102
_ERROR_CODE_MAP: Dict[str, Type[PmxtError]] = {...}
```

Note: `from __future__ import annotations` alone fixes lines 12, 66, 84 (they're in function signatures). Line 102 is a module-level variable annotation — it is also fixed by `from __future__ import annotations` in Python 3.8+.

---

realfishsam:
Fixes #561

### DIFF
--- a/sdks/python/pmxt/errors.py
+++ b/sdks/python/pmxt/errors.py
@@ -5,6 +5,8 @@ These mirror the error hierarchy in the core TypeScript library,
 enabling users to catch specific error types.
 """
 
+from __future__ import annotations
+
 
 class PmxtError(Exception):
     """Base error class for all pmxt errors."""


realfishsam:
## PR Review: VERIFIED

### What This Does
Adds `from __future__ import annotations` to `errors.py`, making PEP 604 union syntax (`str | None`) and PEP 585 generics (`dict[str, type[...]]`) evaluate lazily at runtime, restoring Python 3.8 compatibility.

### Blast Radius
- Python SDK only (`sdks/python/pmxt/errors.py`)
- No core/TS SDK/OpenAPI impact

### Findings
1. The `from __future__ import annotations` import correctly defers evaluation of all annotations in the file, fixing `str | None` on lines 12, 66, 84 and `dict[str, type[PmxtError]]` on line 102 -- all four 3.10+ constructs are covered by this single import.
2. Consistent with existing usage in `constants.py` which already uses the same import.
3. No functional behavior change -- purely a compatibility fix.

### PMXT Pipeline Check
- Field propagation: N/A
- OpenAPI sync: N/A
- Type safety: OK -- annotations still enforce correct types, just evaluated lazily

### Semver Impact
patch -- fixes a `SyntaxError` on Python 3.8/3.9, no API change

### Risk
None substantive. The `from __future__ import annotations` import is well-understood and has no side effects beyond deferred annotation evaluation.

PR #567, opened 2026-05-24
fix: use Python 3.8-compatible generics in models.py

### LINKED ISSUE #562: [python-types] models.py: bare list[T] syntax in SubscribedAddressSnapshot breaks Python 3.8
## File
`sdks/python/pmxt/models.py`

## Missing Optional + Bare Built-in Generics (Python 3.9+ syntax)

Lines 483–489 in the `SubscribedAddressSnapshot` dataclass use the `list[T]` built-in generic form, which requires Python 3.9+:

| Line | Current | Required for Python 3.8 |
|------|---------|------------------------|
| 483 | `trades: Optional[list[Trade]] = None` | `Optional[List[Trade]]` |
| 485 | `positions: Optional[list[Position]] = None` | `Optional[List[Position]]` |
| 487 | `balances: Optional[list[Balance]] = None` | `Optional[List[Balance]]` |

On Python 3.8, the dataclass decorator evaluates field annotations at class-definition time. `list[Trade]` raises:
```
TypeError: 'type' object is not subscriptable
```
immediately on `import pmxt`.

## Inconsistency

Every other collection type in `models.py` correctly uses `List[T]` from `typing` (e.g., `outcomes: List[MarketOutcome]`, `tags: Optional[List[str]]`). These three fields are the only outliers — likely written on a Python 3.9+ machine and not caught because `disallow_untyped_defs = false` and tests presumably ran on Python 3.9+.

## Impact

- **HIGH**: `SubscribedAddressSnapshot` is the public return type of `watch_address()`. Any user on Python 3.8 calling `watch_address()` will receive a crash at import time before the call is even reached.
- This also affects type checkers: `list[Trade]` is not `List[Trade]` in older mypy versions that target Python 3.8 semantics.

## Suggested Fix

```python
from typing import List, Optional  # already imported at top of file

@dataclass
class SubscribedAddressSnapshot:
    address: str
    timestamp: int
    trades: Optional[List[Trade]] = None
    positions: Optional[List[Position]] = None
    balances: Optional[List[Balance]] = None
```

---

realfishsam:
Fixes #562

### DIFF
--- a/sdks/python/pmxt/models.py
+++ b/sdks/python/pmxt/models.py
@@ -480,13 +480,13 @@ class SubscribedAddressSnapshot:
     timestamp: int
 
     """Recent trades for this address"""
-    trades: Optional[list[Trade]] = None
+    trades: Optional[List[Trade]] = None
 
     """Open positions of this address"""
-    positions: Optional[list[Position]] = None
+    positions: Optional[List[Position]] = None
 
     """Balances of this address"""
-    balances: Optional[list[Balance]] = None
+    balances: Optional[List[Balance]] = None
 
 # ----------------------------------------------------------------------------
 # Filtering Types


realfishsam:
## PR Review: VERIFIED

### What This Does
Replaces three `list[T]` (PEP 585, Python 3.9+) annotations with `List[T]` (from `typing`) in the `SubscribedAddressSnapshot` dataclass in `models.py`, restoring Python 3.8 compatibility.

### Blast Radius
- Python SDK only (`sdks/python/pmxt/models.py`)
- No core/TS SDK/OpenAPI impact

### Findings
1. All three occurrences of lowercase `list[T]` in `models.py` are fixed (lines 483, 486, 489). Grep confirms no other instances remain.
2. `List` is already imported from `typing` on line 9 of the file -- no new import needed.
3. The rest of `models.py` already uses `List[T]` and `Optional[...]` consistently; these three were the only outliers.
4. An alternative approach would have been `from __future__ import annotations` (as PR 566 does for `errors.py`), but the `List[T]` approach is also correct and matches the existing style in this file.

### PMXT Pipeline Check
- Field propagation: N/A
- OpenAPI sync: N/A
- Type safety: OK -- semantically identical annotations

### Semver Impact
patch -- fixes `TypeError` on Python 3.8, no API change

### Risk
None. `typing.List[T]` is equivalent to `list[T]` for all supported Python versions.

PR #568, opened 2026-05-24
fix: validate query params in feed-routes instead of unsafe casts

### LINKED ISSUE #558: [non-null] server/feed-routes.ts: 8 unsafe query-param casts at HTTP boundary
## Risk Level
CRITICAL

## File
`core/src/server/feed-routes.ts`

## Findings
- Line 36: `req.query.symbol as string` — passed to `fetchTicker(symbol)`
- Line 49: `req.query.symbols as string | undefined` — then `.split(',')` called on it
- Line 59: `req.query.symbol as string` — passed to `fetchOHLCV(symbol, ...)`
- Line 66: `req.query.timeframe as string` — passed as OHLCV timeframe
- Line 82: `req.query.symbol as string` — passed to `fetchOrderBook(symbol, ...)`
- Line 100: `req.query.feed as string` — passed to `fetchOracleRound({ feed: feedName })`
- Line 118: `req.query.feed as string` — passed to `fetchOracleHistory({ feed: feedName, ... })`
- Line 139: `req.query.symbol as string` — passed to `fetchHistoricalPrices(symbol, ...)`

Express types `req.query.*` as `string | string[] | ParsedQs | ParsedQs[]`. Casting to `string` via `as string` is a TypeScript-only assertion — the runtime value is unchanged. An attacker or misconfigured client can send a repeated query parameter (e.g. `?symbol=a&symbol=b`) and Express will bind `symbol` to `['a', 'b']` (an array). The subsequent null guard `if (!symbol)` passes because `['a', 'b']` is truthy. The array then propagates to exchange methods expecting a plain string, causing `TypeError: symbol.toLowerCase is not a function` or silent corruption deep inside the exchange normalizer.

Line 49 is especially risky: `symbolsRaw.split(',')` called on an array throws immediately.

## Suggested Fix
Validate the actual runtime type before using the value:
```ts
const raw = req.query.symbol;
const symbol = typeof raw === 'string' ? raw : undefined;
if (!symbol) { res.status(400)...; return; }
```
Or use a small helper: `function queryString(v: unknown): string | undefined { return typeof v === 'string' ? v : undefined; }`

---

realfishsam:
Fixes #558

### DIFF
--- a/core/src/server/feed-routes.ts
+++ b/core/src/server/feed-routes.ts
@@ -33,9 +33,9 @@ export function createFeedRouter(): Router {
     // GET /api/feeds/:feed/fetchTicker?symbol=BTC/USD
     router.get('/:feed/fetchTicker', async (req: Request, res: Response, next: NextFunction) => {
         try {
-            const symbol = req.query.symbol as string;
-            if (!symbol) {
-                res.status(400).json({ success: false, error: 'Missing required parameter: symbol' });
+            const symbol = req.query.symbol;
+            if (typeof symbol !== 'string') {
+                res.status(400).json({ success: false, error: 'Missing required query parameter: symbol' });
                 return;
             }
             const data = await (req as any)._feed.fetchTicker(symbol);
@@ -46,7 +46,11 @@ export function createFeedRouter(): Router {
     // GET /api/feeds/:feed/fetchTickers?symbols=BTC/USD,ETH/USD
     router.get('/:feed/fetchTickers', async (req: Request, res: Response, next: NextFunction) => {
         try {
-            const symbolsRaw = req.query.symbols as string | undefined;
+            const symbolsRaw = req.query.symbols;
+            if (symbolsRaw !== undefined && typeof symbolsRaw !== 'string') {
+                res.status(400).json({ success: false, error: 'Invalid query parameter: symbols must be a string' });
+                return;
+            }
             const symbols = symbolsRaw ? symbolsRaw.split(',').map((s) => s.trim()) : undefined;
             const data = await (req as any)._feed.fetchTickers(symbols);
             res.json({ success: true, data });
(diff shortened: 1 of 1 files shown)


realfishsam:
## PR Review: VERIFIED

### What This Does
Replaces unsafe `as string` casts on Express `req.query` values with proper `typeof` runtime checks in all feed-routes handlers. Previously, array query params (e.g., `?symbol=a&symbol=b`) would silently pass as `string[]`, potentially causing downstream bugs. Now they return 400 with a clear error message.

### Blast Radius
- Core server only (`core/src/server/feed-routes.ts`)
- Affects feed endpoints: fetchTicker, fetchTickers, fetchOHLCV, fetchOrderBook, fetchOracleRound, fetchOracleHistory, fetchHistoricalPrices
- No exchange/SDK/OpenAPI impact

### Findings
1. The validation pattern is correct: `typeof x !== 'string'` catches both missing (`undefined`) and array (`string[]`) cases. Required params reject all non-string values; optional params check `!== undefined` first.
2. The `fetchHistoricalPrices` handler still has a residual `as 'asc' | 'desc' | undefined` cast on line ~163 of the diff. This is acceptable -- at that point `order` has already been validated as `string | undefined` by the guard above, so the cast is narrowing a verified string to the union type. Not unsafe.
3. Line 17 still has `req.params.feed as string` -- this is fine since Express route params are always strings (unlike query params).
4. Error messages are consistently updated to say "query parameter" which is more precise than the previous "parameter".

### PMXT Pipeline Check
- Field propagation: N/A
- OpenAPI sync: N/A
- Type safety: OK -- removes unsafe casts, adds runtime validation

### Semver Impact
patch -- tightens input validation, no new functionality. Could technically break clients sending malformed array query params, but those were already producing undefined behavior.

### Risk
Low. The `Number()` coercions on `since`, `limit`, `fromTimestamp`, `untilTimestamp`, `maxSize` remain unvalidated (e.g., `Number("abc")` produces `NaN`). This is pre-existing and out of scope for this fix but worth noting for a follow-up.

PR #569, opened 2026-05-24
fix: validate parsed.method in ws-handler before use

### LINKED ISSUE #559: [non-null] server/ws-handler.ts: parsed.method cast to string without validation
## Risk Level
CRITICAL

## File
`core/src/server/ws-handler.ts`

## Findings
- Line 270: `method: parsed.method as string` — inside `SubscribeMessage` construction
- Line 280: `method: parsed.method as string` — inside `UnsubscribeMessage` construction

The required-fields guard on line 258 checks only `id`, `action`, and `exchange`:
```ts
if (!id || !action || !exchange) {
  sendError(ws, id, "Missing required fields: id, action, exchange");
  return;
}
```
`method` is not validated. A WebSocket client that sends:
```json
{"id":"1","action":"subscribe","exchange":"kalshi"}
```
passes the guard. `parsed.method` is `undefined` at runtime; `as string` is a compile-time fiction. The `undefined` value is stored as `msg.method` in `SubscribeMessage`. When the subscription handler then executes `(exchange as any)[method]` (line 373), it evaluates `exchange[undefined]`, which is `undefined`, triggering `TypeError: undefined is not a function`.

A client can also inject non-string JSON values (e.g. `"method": {"toString": ...}`) that survive the cast and cause unexpected behavior when the method name is string-interpolated into error messages or log lines.

## Suggested Fix
Add `method` to the required-fields check, or validate it explicitly:
```ts
const method = parsed.method;
if (typeof method !== 'string' || !method) {
  sendError(ws, id, "Missing required field: method");
  return;
}
```

---

realfishsam:
Fixes #559

### DIFF
--- a/core/src/server/ws-handler.ts
+++ b/core/src/server/ws-handler.ts
@@ -255,8 +255,10 @@ export function createWebSocketHandler(
       const action = parsed.action as string | undefined;
       const exchange = parsed.exchange as string | undefined;
 
-      if (!id || !action || !exchange) {
-        sendError(ws, id, "Missing required fields: id, action, exchange");
+      const method = parsed.method as string | undefined;
+
+      if (!id || !action || !exchange || !method) {
+        sendError(ws, id, "Missing required fields: id, action, exchange, method");
         return;
       }
 
@@ -267,7 +269,7 @@ export function createWebSocketHandler(
           id,
           action: "subscribe",
           exchange: exchangeName,
-          method: parsed.method as string,
+          method,
           args: (parsed.args as unknown[]) || [],
           credentials: parsed.credentials as ExchangeCredentials | undefined,
         };
@@ -277,7 +279,7 @@ export function createWebSocketHandler(
           id,
           action: "unsubscribe",
           exchange: exchangeName,
-          method: parsed.method as string,
+          method,
           args: (parsed.args as unknown[]) || [],
         };
         handleUnsubscribe(ws, state, msg, exchangeName);


realfishsam:
## PR Review: VERIFIED

### What This Does
Validates `parsed.method` from WebSocket messages before use, promoting it from an unvalidated `parsed.method as string` cast to a guarded extraction with an early-return error. Previously a client could omit `method` and the value `undefined` would be silently cast to `string`, propagating through subscribe/unsubscribe handlers.

### Blast Radius
- Core server only (`core/src/server/ws-handler.ts`)
- Affects all WebSocket subscribe/unsubscribe message handling
- No exchange/SDK/OpenAPI impact

### Findings
1. The `method` field is already required in both `SubscribeMessage` and `UnsubscribeMessage` interfaces (lines 12-27), so validating it at the parsing boundary is correct and consistent with the type contract.
2. The guard `!method` correctly catches both `undefined` (missing field) and `""` (empty string). The error message is updated to list all four required fields.
3. After the guard, `method` is a narrowed `string`, so the downstream assignments (`method: method`) no longer need `as string` casts. Clean.
4. The extraction pattern (`const method = parsed.method as string | undefined`) is consistent with how `id`, `action`, and `exchange` are extracted on lines 254-256.

### PMXT Pipeline Check
- Field propagation: N/A
- OpenAPI sync: N/A
- Type safety: OK -- eliminates unsafe cast, adds runtime validation

### Semver Impact
patch -- adds input validation to WebSocket handler. Clients that were omitting `method` were already getting undefined behavior; now they get a clear error.

### Risk
None. The only behavioral change is that malformed messages missing `method` now get an explicit error instead of silently failing downstream.

PR #570, opened 2026-05-24
fix: replace non-null assertions with guard checks in limitless

### LINKED ISSUE #560: [non-null] limitless/index.ts: 9 unsafe assertions on optional interface methods
## Risk Level
MEDIUM

## File
`core/src/exchanges/limitless/index.ts`

## Findings
- Line 200: `this.fetcher.fetchRawOHLCV!(slug, params)` — method is `fetchRawOHLCV?` in `IExchangeFetcher`
- Line 201: `this.normalizer.normalizeOHLCV!(rawPrices as any, params)` — method is `normalizeOHLCV?` in `IExchangeNormalizer`
- Line 209: `this.fetcher.fetchRawOrderBook!(slug)` — method is `fetchRawOrderBook?` in `IExchangeFetcher`
- Line 210: `this.normalizer.normalizeOrderBook!(rawOrderBook as any, outcomeId)` — method is `normalizeOrderBook?` in `IExchangeNormalizer`
- Line 254: `this.fetcher.fetchRawTrades!(slug, params)` — method is `fetchRawTrades?` in `IExchangeFetcher`
- Line 255: `this.normalizer.normalizeTrade!(raw, i)` — method is `normalizeTrade?` in `IExchangeNormalizer`
- Line 260: `this.fetcher.fetchRawMyTrades!(params || {}, auth.getApiKey())` — method is `fetchRawMyTrades?` in `IExchangeFetcher`
- Line 261: `this.normalizer.normalizeUserTrade!(raw, i)` — method is `normalizeUserTrade?` in `IExchangeNormalizer`
- Line 441: `this.normalizer.normalizePosition!(raw)` — method is `normalizePosition?` in `IExchangeNormalizer`

All nine methods are declared optional (with `?`) in `IExchangeFetcher` / `IExchangeNormalizer` (`core/src/exchanges/interfaces.ts` lines 29–33, 46–50). The concrete `LimitlessFetcher` and `LimitlessNormalizer` currently implement them all, so at runtime with the default classes the assertions hold.

However, the `!` silences TypeScript's check entirely. If a consumer:
- Passes a custom fetcher or normalizer that skips any of these methods
- Replaces `LimitlessFetcher` with a mock in tests without all methods
- Future refactor removes or renames any of these methods

...the `!` ensures TypeScript stays silent while the runtime throws `TypeError: this.fetcher.fetchRawOHLCV is not a function`.

`normalizer.ts` line 104–106 already documents that `normalizeTrade` is not supported and throws; `fetcher.fetchRawTrades!` calling into it will always throw at runtime — the `!` hides this known-broken path.

## Suggested Fix
Guard each call with a capability check before calling, or replace the optional interface methods with concrete required methods in the Limitless-specific types:
```ts
if (!this.fetcher.fetchRawOHLCV) throw new Error('Limitless fetcher does not support OHLCV');
const rawPrices = await this.fetcher.fetchRawOHLCV(slug, params);
```

---

realfishsam:
Fixes #560

### DIFF
--- a/core/src/exchanges/limitless/index.ts
+++ b/core/src/exchanges/limitless/index.ts
@@ -197,8 +197,10 @@ export class LimitlessExchange extends PredictionMarketExchange {
 
     async fetchOHLCV(outcomeId: string, params: OHLCVParams): Promise<PriceCandle[]> {
         const slug = await this.resolveSlug(outcomeId);
-        const rawPrices = await this.fetcher.fetchRawOHLCV!(slug, params);
-        return this.normalizer.normalizeOHLCV!(rawPrices as any, params);
+        if (!this.fetcher.fetchRawOHLCV) { throw new Error('fetchRawOHLCV is not implemented for this exchange'); }
+        const rawPrices = await this.fetcher.fetchRawOHLCV(slug, params);
+        if (!this.normalizer.normalizeOHLCV) { throw new Error('normalizeOHLCV is not implemented for this exchange'); }
+        return this.normalizer.normalizeOHLCV(rawPrices as any, params);
     }
 
     async fetchOrderBook(outcomeId: string, limit?: number, params?: Record<string, any>): Promise<OrderBook> {
@@ -206,8 +208,10 @@ export class LimitlessExchange extends PredictionMarketExchange {
         outcomeId = resolved.outcomeId;
         params = resolved.params;
         const slug = await this.resolveSlug(outcomeId);
-        const rawOrderBook = await this.fetcher.fetchRawOrderBook!(slug);
-        const orderBook = this.normalizer.normalizeOrderBook!(rawOrderBook as any, outcomeId);
+        if (!this.fetcher.fetchRawOrderBook) { throw new Error('fetchRawOrderBook is not implemented for this exchange'); }
+        const rawOrderBook = await this.fetcher.fetchRawOrderBook(slug);
+        if (!this.normalizer.normalizeOrderBook) { throw new Error('normalizeOrderBook is not implemented for this exchange'); }
+        const orderBook = this.normalizer.normalizeOrderBook(rawOrderBook as any, outcomeId);
 
         // The Limitless API always returns the Yes-side order book regardless
         // of which token is queried. If the caller asked for the No token,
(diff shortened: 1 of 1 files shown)


realfishsam:
## PR Review: VERIFIED

### What This Does
Replaces non-null assertions (`!`) with explicit guard-and-throw checks in the Limitless exchange class for optional fetcher/normalizer methods. The methods (`fetchRawOHLCV`, `fetchRawOrderBook`, `fetchRawTrades`, `fetchRawMyTrades`, `normalizeOHLCV`, `normalizeOrderBook`, `normalizeTrade`, `normalizeUserTrade`, `normalizePosition`) are declared as optional in the `ExchangeFetcher`/`ExchangeNormalizer` interfaces, so the non-null assertions were suppressing real type-safety.

### Blast Radius
- Core exchange: Limitless only (`core/src/exchanges/limitless/index.ts`)
- Affects: fetchOHLCV, fetchOrderBook, fetchTrades, fetchMyTrades, fetchPositions
- No SDK/OpenAPI/other exchange impact

### Findings
1. All 9 non-null assertions are replaced with proper guards. Error messages are descriptive ("fetchRawOHLCV is not implemented for this exchange").
2. For `fetchTrades` and `fetchMyTrades`, the normalizer method is captured to a local variable before the `.map()` call (`const normalizeTrade = this.normalizer.normalizeTrade`). This is necessary because TypeScript cannot narrow optional methods inside callback closures. Correct pattern.
3. For `fetchPositions`, the same local-variable capture pattern is used for `normalizePosition`. Consistent.
4. The `fetchRawPositions` call on line 440 does NOT get a guard -- this is correct because `fetchRawPositions` is a non-optional method in the interface (it's always implemented).
5. The Limitless fetcher/normalizer actually implement all these methods (confirmed via grep of `fetcher.ts` and `normalizer.ts`), so the guards will never fire at runtime today. But they provide correct type-safety if the interface contracts change.

### PMXT Pipeline Check
- Field propagation: N/A
- OpenAPI sync: N/A
- Type safety: OK -- strictly improves type safety by replacing compile-time suppression with runtime validation

### Semver Impact
patch -- no API change. The new `Error` throws replace what would have been `TypeError: undefined is not a function` at runtime.

### Risk
Low. The guards are technically dead code today (Limitless implements all these methods), but they provide correct defense against future interface changes. No behavioral change for working calls.

PR #571, opened 2026-05-24
perf: replace O(n²) concat with push in metaculus fetchMarkets

### LINKED ISSUE #551: Unbounded: core/src/exchanges/metaculus/fetchMarkets.ts — O(n²) array.concat() with MAX_PAGES=200
## Location
`core/src/exchanges/metaculus/fetchMarkets.ts:59`

## Code
```typescript
const BATCH_SIZE = 100;
const MAX_PAGES = 200; // safety cap (~20 000 posts)

async function fetchPostPages(
    callApi: CallApi,
    apiParams: Record<string, any>,
    targetCount?: number,
): Promise<any[]> {
    let all: any[] = [];
    let offset = 0;
    let page = 0;

    do {
        const data = await callApi("GetPosts", { ...apiParams, limit: BATCH_SIZE, offset });
        const results: any[] = data.results ?? [];
        if (results.length === 0) break;

        all = all.concat(results);  // ← creates a new array every iteration
        offset += results.length;
        page++;

        if (targetCount && all.length >= targetCount * 1.5) break;
        if (!data.next) break;
    } while (page < MAX_PAGES);

    return all;
}
```

## Growth Pattern
`Array.prototype.concat()` allocates a brand-new array on every call — it does not mutate in place. At page `k`, the result array has `k × BATCH_SIZE` slots and the previous array (same size minus one page) is eligible for GC but may not be collected before the next iteration. Both arrays coexist in memory simultaneously.

Total pointer slots allocated across all iterations:
`BATCH_SIZE × (1 + 2 + … + MAX_PAGES) = 100 × (200 × 201 / 2)` = **2,010,000** pointer copies

`fetchPostPages` is called from multiple code paths in `fetchMarkets.ts`:
- `fetchMarketsDefault` with `fetchLimit = 2000` for volume/liquidity sorts
- `searchMarkets` with `Math.max(limit * 5, 500)` posts
- Slug fallback path with `500` posts
- `eventId` path with up to `params.limit ?? 1000` posts

## OOM Estimate
- At `MAX_PAGES=200`, `BATCH_SIZE=100`: ~2M pointer slots → **~16 MB** of intermediate arrays per full `fetchMarkets()` call
- Each Metaculus post object: ~2–5 KB (nested questions, forecasts, projects)
- Full 20,000-post crawl: ~40–100 MB of live data + 16 MB of GC pressure
- Under 5 concurrent `fetchMarkets()` calls (e.g., server handling simultaneous client requests): **~80 MB** of short-lived garbage in flight simultaneously, causing GC pauses

## Suggested Fix
Replace `all = all.concat(results)` with `all.push(...results)`. Since `BATCH_SIZE=100` is small, spreading onto the call stack is safe. This eliminates all intermediate array allocations, reducing the pattern from O(n²) allocations to O(n).

```typescript
all.push(...results); // in-place, no intermediate array
```

---

realfishsam:
Fixes #551

### DIFF
--- a/core/src/exchanges/metaculus/fetchMarkets.ts
+++ b/core/src/exchanges/metaculus/fetchMarkets.ts
@@ -56,7 +56,7 @@ async function fetchPostPages(
         const results: any[] = data.results ?? [];
         if (results.length === 0) break;
 
-        all = all.concat(results);
+        all.push(...results);
         offset += results.length;
         page++;
 


realfishsam:

### What This Does
Replaces `all = all.concat(results)` with `all.push(...results)` in the metaculus `fetchMarkets.ts` pagination loop, avoiding O(n^2) array re-allocation on each page.

### Blast Radius
- **Core**: `core/src/exchanges/metaculus/fetchMarkets.ts` only

### Findings
1. The `all` variable is declared with `let` (line 45) but is no longer reassigned after this change. It could be tightened to `const` for clarity, though this is non-blocking.
2. With `push(...results)` and `BATCH_SIZE = 100`, each spread is well under the V8 call-stack limit (~65k args). No risk of "Maximum call stack size exceeded" at this batch size.
3. Correct behavior preserved: `all.length` continues to reflect cumulative results for the early-exit check on line 64.

### PMXT Pipeline Check
- Field propagation: N/A
- OpenAPI sync: N/A
- Type safety: OK

### Semver Impact
patch -- internal perf improvement, no API or behavioral change

### Risk
Not verified against live Metaculus data. The change is mechanically safe but untested in this review.

PR #572, opened 2026-05-24
perf: replace O(n²) concat with push in metaculus fetchEvents

### LINKED ISSUE #552: Unbounded: core/src/exchanges/metaculus/fetchEvents.ts — O(n²) array.concat() with MAX_PAGES=200
## Location
`core/src/exchanges/metaculus/fetchEvents.ts:46`

## Code
```typescript
const BATCH_SIZE = 100;
const MAX_PAGES = 200;

async function fetchPostPages(
    callApi: CallApi,
    apiParams: Record<string, any>,
    targetCount?: number,
): Promise<any[]> {
    let all: any[] = [];
    let offset = 0;
    let page = 0;

    do {
        const data = await callApi("GetPosts", { ...apiParams, limit: BATCH_SIZE, offset });
        const results: any[] = data.results ?? [];
        if (results.length === 0) break;

        all = all.concat(results);  // ← creates a new array every iteration
        offset += results.length;
        page++;

        if (targetCount && all.length >= targetCount) break;
        if (!data.next) break;
    } while (page < MAX_PAGES);

    return all;
}
```

## Growth Pattern
This is the same `fetchPostPages` function duplicated verbatim in `fetchEvents.ts` (the one in `fetchMarkets.ts` is covered by a separate issue). `Array.prototype.concat()` allocates a new array on every call — at page `k`, the array has `k × BATCH_SIZE` elements and the old array coexists in memory until GC.

`fetchPostPages` is called in `fetchEvents.ts` from three code paths:
- `fetchEventBySlug` (tournament slug path): up to 100 posts
- `fetchEventBySlug` (slug-string match fallback): up to 500 posts
- Default listing path `fetchEvents`: up to `(offset + limit) × 5` posts

The slug fallback at line 159 is especially risky:
```typescript
const posts = await fetchPostPages(
    callApi,
    { with_cp: true, order_by: "-forecasts_count" },
    500,  // fetches 500 posts to find one matching slug
);
```
This is called on every unrecognized slug and can fetch up to 500 posts via concat.

## OOM Estimate
- Full 200-page crawl: `100 × (200 × 201 / 2)` = 2,010,000 pointer copies → **~16 MB** intermediate
- Per-symbol slug lookup (500 posts): `100 × (5 × 6 / 2)` = 1,500 pointer copies → negligible per call, but if N concurrent slug lookups occur: N × 1.5 MB overhead
- Under 10 concurrent `fetchEvents()` calls with `sort='volume'` (fetching all pages): **~160 MB** of short-lived garbage

## Suggested Fix
Replace `all = all.concat(results)` with `all.push(...results)`. `BATCH_SIZE=100` is safe to spread onto the call stack. This is the same fix as the `fetchMarkets.ts` counterpart — the two functions are identical and could be deduplicated into a shared utility.

---

realfishsam:
Fixes #552

### DIFF
--- a/core/src/exchanges/metaculus/fetchEvents.ts
+++ b/core/src/exchanges/metaculus/fetchEvents.ts
@@ -42,7 +42,7 @@ async function fetchPostPages(
         const results: any[] = data.results ?? [];
         if (results.length === 0) break;
 
-        all = all.concat(results);
+        all.push(...results);
         offset += results.length;
         page++;
 


realfishsam:

### What This Does
Replaces `all = all.concat(results)` with `all.push(...results)` in the metaculus `fetchEvents.ts` pagination loop, mirroring the same O(n^2) fix applied in PR #571.

### Blast Radius
- **Core**: `core/src/exchanges/metaculus/fetchEvents.ts` only

### Findings
1. Same `let` -> could-be-`const` note as PR #571 (line 31). Non-blocking.
2. Batch size is 100, safely below V8 spread argument limits.
3. Both fetchMarkets and fetchEvents share the same `fetchPostPages` pattern. If these were ever consolidated into a shared helper, only one fix would be needed. Minor tech-debt observation, not blocking.

### PMXT Pipeline Check
- Field propagation: N/A
- OpenAPI sync: N/A
- Type safety: OK

### Semver Impact
patch -- internal perf improvement, no API or behavioral change

### Risk
Not verified against live Metaculus data. Mechanically identical to PR #571.

PR #573, opened 2026-05-24
fix: clean up orderBookResolvers after resolution in probable websocket

### LINKED ISSUE #550: Unbounded: core/src/exchanges/probable/websocket.ts — orderBookResolvers leak on watchTimeout
## Location
`core/src/exchanges/probable/websocket.ts:73-84`

## Code
```typescript
async watchOrderBook(tokenId: string): Promise<OrderBook> {
    const client = await this.ensureClient();

    if (!this.subscriptions.has(tokenId)) {
        const sub = client.subscribePublicStream(
            [`book:${tokenId}`],
            (data: any) => { this.handleOrderBookUpdate(tokenId, data); }
        );
        this.subscriptions.set(tokenId, sub);
    }

    // Push resolver into array — never removed if timeout wins the race
    const dataPromise = new Promise<OrderBook>((resolve, reject) => {
        if (!this.orderBookResolvers.has(tokenId)) {
            this.orderBookResolvers.set(tokenId, []);
        }
        this.orderBookResolvers.get(tokenId)!.push({ resolve, reject }); // ← stale on timeout
    });

    return withWatchTimeout(
        dataPromise,
        this.config.watchTimeoutMs ?? DEFAULT_WATCH_TIMEOUT_MS,
        `watchOrderBook('${tokenId}')`,
    );
}
```

`resolveOrderBook` (line 116) clears the array only when a WS update fires:
```typescript
private resolveOrderBook(tokenId: string, orderBook: OrderBook) {
    const resolvers = this.orderBookResolvers.get(tokenId);
    if (resolvers && resolvers.length > 0) {
        resolvers.forEach(r => r.resolve(orderBook));
        this.orderBookResolvers.set(tokenId, []); // ← only cleared on WS update
    }
}
```

## Growth Pattern
`withWatchTimeout` races `dataPromise` against a timer. When the **timeout wins** (sparse market, WS not yet delivering), `dataPromise` is abandoned but its `{ resolve, reject }` closure is still referenced inside `orderBookResolvers.get(tokenId)`. The next call to `watchOrderBook` pushes another closure. Over time the array accumulates one stale closure per timed-out call.

This is the exact same structural defect as issue #372 (Limitless websocket) — timeout wins the race but doesn't clean up its resolver.

## OOM Estimate
- Each `{ resolve, reject }` closure: ~1–2 KB of V8 heap (captured Promise internals)
- CCXT Pro loop at 1 call/sec, 30-second timeout for a sparse market: up to 30 stale closures per WS update cycle
- 100 active tokenIds × 30 closures × 2 KB = **6 MB** permanently held between WS updates
- Over a 24-hour session with occasional WS gaps (e.g., reconnects): hundreds of closures accumulate per token — **50–200 MB**

## Suggested Fix
In the timeout handler inside `withWatchTimeout` (or in `watchOrderBook` itself), remove the specific `{ resolve, reject }` from `orderBookResolvers` before letting the timeout resolve. Additionally, add a `close()` path that calls `this.orderBookResolvers.clear()` and rejects all pending resolvers.

---

realfishsam:
Fixes #550

### DIFF
--- a/core/src/exchanges/probable/websocket.ts
+++ b/core/src/exchanges/probable/websocket.ts
@@ -70,15 +70,25 @@ export class ProbableWebSocket {
         }
 
         // Return a promise that resolves on the next orderbook update
+        let resolver: QueuedPromise<OrderBook>;
         const dataPromise = new Promise<OrderBook>((resolve, reject) => {
-            this.getOrderBookQueue(tokenId).push({ resolve, reject });
+            resolver = { resolve, reject };
+            this.getOrderBookQueue(tokenId).push(resolver);
         });
 
         return withWatchTimeout(
             dataPromise,
             this.config.watchTimeoutMs ?? DEFAULT_WATCH_TIMEOUT_MS,
             `watchOrderBook('${tokenId}')`,
-        );
+        ).finally(() => {
+            const queue = this.orderBookResolvers.get(tokenId);
+            if (queue) {
+                const idx = queue.indexOf(resolver!);
+                if (idx !== -1) {
+                    queue.splice(idx, 1);
+                }
+            }
+        });
     }
 
     private handleOrderBookUpdate(tokenId: string, data: any) {


realfishsam:

### What This Does
Adds a `.finally()` handler to the `watchOrderBook` promise in the Probable WebSocket that removes the resolver from the queue after the promise settles (whether by data arrival or timeout). This prevents unbounded growth of the `orderBookResolvers` map when timeouts occur.

### Blast Radius
- **Core**: `core/src/exchanges/probable/websocket.ts` only

### Findings
1. The `resolver` variable is declared with `let` (uninitialized) and then assigned inside the Promise constructor. The `resolver!` non-null assertion on line where `indexOf` is called is technically safe because the `.finally()` can only fire after the Promise constructor has run, but it relies on a subtle execution order guarantee. A cleaner pattern would be to initialize `resolver` inline or use a sentinel, though this is non-blocking.
2. **Potential double-removal is harmless**: `resolveOrderBook()` (line 112-118) already clears the entire queue via `this.orderBookResolvers.set(tokenId, [])` on data arrival. The `.finally()` then calls `indexOf` on the fresh empty array and finds nothing -- no bug, just a no-op.
3. **splice mutates the array in-place** -- this is acceptable here because the queue array is an internal mutable structure, not user-facing data.
4. The `close()` method (line 131) already rejects all pending resolvers and clears the map, so `.finally()` cleanup interacts safely with shutdown.

### PMXT Pipeline Check
- Field propagation: N/A
- OpenAPI sync: N/A
- Type safety: OK (the `resolver!` assertion is safe given execution semantics)

### Semver Impact
patch -- internal memory leak fix, no API or behavioral change

### Risk
Not verified with a live Probable WebSocket connection. The interaction between `resolveOrderBook` (replaces entire array) and the `.finally()` (splices individual entry) is benign but worth noting in case the resolution strategy changes later.

PR #574, opened 2026-05-24
fix: clean up subscriptions and dataStore on unsubscribe in ws-client

### LINKED ISSUE #553: Unbounded: sdks/typescript/pmxt/ws-client.ts — subscriptions and dataStore never evicted in subscribeBatch
## Location
`sdks/typescript/pmxt/ws-client.ts:248-304` (`subscribeBatch`) and lines 43-46 (the Maps)

## Code
```typescript
// Instance-level Maps — grow without bound during session lifetime
private dataStore: Map<string, any> = new Map();      // grows forever
private subscriptions: Map<string, WsSubscription> = new Map(); // grows forever
// (activeSubs Map is only used by subscribe(), not subscribeBatch())

async subscribeBatch(
    exchange: string,
    method: string,
    args: any[],
    credentials?: Record<string, any>,
    timeoutMs = 30000,
): Promise<Record<string, any>> {
    const symbols: string[] = Array.isArray(args[0]) ? args[0] : [];

    await this.ensureConnected();

    const requestId = `req-${Math.random().toString(36).slice(2, 14)}`; // ← fresh ID every call

    const sub: WsSubscription = {
        requestId,
        method,
        symbols,
        resolve: null,
        reject: null,
    };
    this.subscriptions.set(requestId, sub);  // ← never removed

    this.ws!.send(JSON.stringify(message));

    await this.waitForData(requestId, timeoutMs);

    // Collect per-symbol data — entries never deleted afterward
    const result: Record<string, any> = {};
    for (const symbol of symbols) {
        const storeKey = `${requestId}:${symbol}`;
        const data = this.dataStore.get(storeKey); // ← read but never deleted
        if (data !== undefined) {
            result[symbol] = data;
        }
    }
    return result;
}
```

## Growth Pattern
Unlike `subscribe()`, which deduplicates via `activeSubs` and reuses the same `requestId` for the same symbol set, `subscribeBatch()` generates a fresh UUID on **every call**. This means:

1. `this.subscriptions[requestId]` gets a new entry per `subscribeBatch()` call, never removed.
2. `this.dataStore["${requestId}:${symbol}"]` gets a new entry per symbol per call, never deleted after being read.
3. `this.dataStore[requestId]` (the fallback at line 303) also persists.

`subscribeBatch` is the backing transport for `watchOrderBooks()` in the TypeScript SDK. In the CCXT Pro pattern, users call `watchOrderBooks()` in a tight `while (true)` loop — one call per desired tick. Each iteration silently leaks entries.

This is the exact TypeScript analog of Python issue #363.

## OOM Estimate
- Per `subscribeBatch()` call: 1 `WsSubscription` object (~200 bytes) + N dataStore entries (each an order book dict, ~500 bytes per symbol)
- Watching 50 symbols, polling at 1 Hz:
  - After 1 hour: 3,600 subscription entries + 180,000 dataStore entries
  - Memory: 3,600 × 200 B + 180,000 × 500 B = **~90 MB** after 1 hour
- After 24 hours: **~2.2 GB** — OOM on typical Node.js deployments with default heap limits
- V8 GC cannot collect these entries because `this.subscriptions` and `this.dataStore` hold live references

## Suggested Fix
After `subscribeBatch()` collects its result, delete the consumed entries:

```typescript
// After collecting result:
this.subscriptions.delete(requestId);
for (const symbol of symbols) {
    this.dataStore.delete(`${requestId}:${symbol}`);
}
this.dataStore.delete(requestId);
```

Also add deduplication via `activeSubs` (same as `subscribe()`) so repeated batch calls for the same symbol set reuse the existing subscription instead of creating a new one each time.

---

realfishsam:
Fixes #553

### DIFF
--- a/sdks/typescript/pmxt/ws-client.ts
+++ b/sdks/typescript/pmxt/ws-client.ts
@@ -254,6 +254,14 @@ export class SidecarWsClient {
     ): Promise<Record<string, any>> {
         const symbols: string[] = Array.isArray(args[0]) ? args[0] : [];
 
+        // Deduplicate via activeSubs -- reuse existing subscription for the same symbol set
+        const subKey = `${method}:${[...symbols].sort().join(",")}`;
+        const existingId = this.activeSubs.get(subKey);
+        if (existingId && this.subscriptions.has(existingId)) {
+            await this.waitForData(existingId, timeoutMs);
+            return this.collectBatchResult(existingId, symbols);
+        }
+
         await this.ensureConnected();
 
         const requestId = `req-${Math.random().toString(36).slice(2, 14)}`;
@@ -266,6 +274,7 @@ export class SidecarWsClient {
             reject: null,
         };
         this.subscriptions.set(requestId, sub);
+        this.activeSubs.set(subKey, requestId);
 
         const message: Record<string, any> = {
             id: requestId,
@@ -283,19 +292,31 @@ export class SidecarWsClient {
         // Wait for first data event
         await this.waitForData(requestId, timeoutMs);
 
-        // Collect per-symbol data
+        return this.collectBatchResult(requestId, symbols);
+    }
+
+    /**
+     * Collect per-symbol data for a batch subscription and clean up
+     * consumed dataStore entries to prevent unbounded growth.
+     */
+    private collectBatchResult(
+        requestId: string,
+        symbols: readonly string[],
+    ): Record<string, any> {
         const result: Record<string, any> = {};
         for (const symbol of symbols) {
             const storeKey = `${requestId}:${symbol}`;
             const data = this.dataStore.get(storeKey);
             if (data !== undefined) {
                 result[symbol] = data;
+                this.dataStore.delete(storeKey);
             }
         }
 
         // If no per-symbol data, return the single data event as-is
         if (Object.keys(result).length === 0) {
             const data = this.dataStore.get(requestId);
+            this.dataStore.delete(requestId);
             if (data && typeof data === "object") {
                 return data;
             }
@@ -306,6 +327,9 @@ export class SidecarWsClient {
 
     close(): void {
         this.closed = true;
+        this.subscriptions.clear();
+        this.dataStore.clear();
+        this.activeSubs.clear();
         if (this.ws) {
             try {
                 this.ws.close();


realfishsam:
## PR Review: FAIL

### What This Does
Adds subscription deduplication to `subscribeBatch`, cleans up `dataStore` entries after collection, and clears all maps on `close()` in the TS SDK WebSocket client. Targets unbounded memory growth from repeated batch subscriptions.

### Blast Radius
- **SDK**: `sdks/typescript/pmxt/ws-client.ts` only
- No core, OpenAPI, or Python SDK changes

### Findings
1. **BLOCKING -- `subscribeBatch` dedup creates stale-data bug**: The dedup block (lines added at top of `subscribeBatch`) reuses an existing subscription and calls `this.waitForData(existingId, timeoutMs)`. But `waitForData` (line 327-358) checks `this.dataStore.get(requestId)` and if data exists, immediately resolves and *deletes* it. Then `collectBatchResult` is called, which also tries `this.dataStore.get(storeKey)` -- but the data was already deleted by `waitForData`. The result will be an empty object `{}` on the second call to the same subscription, losing all data. This is a silent data-loss bug.

2. **BLOCKING -- `collectBatchResult` deletes data that `subscribe` (non-batch) still needs**: The `subscribe` method on main (line 199-246) already has its own `activeSubs` dedup and `waitForData` flow. But `subscribeBatch` now deletes from `dataStore` in `collectBatchResult`. If `subscribe` and `subscribeBatch` ever share a requestId (they don't today, but `activeSubs` keys could collide if the same method+symbols are used), the deletion would cause data loss in the other path.

3. **`close()` cleanup is correct**: Clearing `subscriptions`, `dataStore`, and `activeSubs` on close is a good practice and prevents memory leaks on reconnection.

4. **`collectBatchResult` extraction is clean**: The refactor to extract the collection logic into a helper is a good structural improvement.

### PMXT Pipeline Check
- Field propagation: N/A
- OpenAPI sync: N/A
- Type safety: OK

### Semver Impact
patch -- internal fix, no public API change

### Risk
Finding #1 is a data-loss regression. When `subscribeBatch` is called twice for the same symbol set, the second call will return an empty object because `waitForData` consumes the data before `collectBatchResult` can read it. This needs to be fixed before merge -- either `waitForData` should not delete data when called from the dedup path, or the dedup path should skip `waitForData` and go directly to `collectBatchResult` with freshness checks.

PR #575, opened 2026-05-24
fix: add 30s timeout to GoldSky subscriber fetch

### LINKED ISSUE #512: Missing timeout: GoldSky subscriber — fetch() has AbortSignal but no time-based abort
## Location
`core/src/subscriber/external/goldsky.ts:350`

## Code
```typescript
// AbortController created at line 333 — never given a time-based abort
const res = await fetch(q.url, {
    method: 'POST',
    headers,
    body: JSON.stringify({ query: q.query, variables: q.variables ?? {} }),
    signal,  // ← controller.signal — only aborted on next poll or unsubscribe
});
```

## Risk
`runQuery()` passes an `AbortController.signal` to `fetch`, which looks safe but isn't: the controller is only aborted when the subscriber unsubscribes or a new poll cycle starts — not on elapsed time. If the GoldSky GraphQL endpoint hangs (slow response, TCP stall), the fetch will block until the *next* poll interval fires and the previous signal gets aborted. In practice this means a single hung query can stall activity subscription data for the full poll period with no error surfaced.

## Affected Methods
- `GoldSkySubscriber.runQuery()` — called for all on-chain activity subscriptions
- Any consumer awaiting GoldSky activity events (trade history, settlement data)

## Suggested Fix
Replace the plain `AbortController` with a time-bounded signal, either:

```typescript
// Option A — merge timeout + cancellation signals
const timeoutSignal = AbortSignal.timeout(15_000);
const res = await fetch(q.url, {
    method: 'POST',
    headers,
    body: JSON.stringify({ query: q.query, variables: q.variables ?? {} }),
    signal: AbortSignal.any([signal, timeoutSignal]),
});
```

```typescript
// Option B — setTimeout on existing controller
const timeoutId = setTimeout(() => controller.abort(new Error('GoldSky fetch timeout')), 15_000);
try {
    const res = await fetch(q.url, { ..., signal });
} finally {
    clearTimeout(timeoutId);
}
```

---
*Found by automated missing timeout audit*

realfishsam:
Fixes #512

### DIFF
--- a/core/src/subscriber/external/goldsky.ts
+++ b/core/src/subscriber/external/goldsky.ts
@@ -346,12 +346,18 @@ export class GoldSkySubscriber implements BaseSubscriber {
             headers['Authorization'] = `Bearer ${this.config.apiKey}`;
         }
 
+        const timeoutSignal = AbortSignal.timeout(30_000);
+        const combinedController = new AbortController();
+        const onAbort = () => combinedController.abort();
+        signal.addEventListener('abort', onAbort, { once: true });
+        timeoutSignal.addEventListener('abort', onAbort, { once: true });
+
         try {
             const res = await fetch(q.url, {
                 method: 'POST',
                 headers,
                 body: JSON.stringify({ query: q.query, variables: q.variables ?? {} }),
-                signal,
+                signal: combinedController.signal,
             });
             if (!res.ok) {
                 logger.warn(`GoldSkySubscriber: HTTP ${res.status} from ${q.url}`);
@@ -364,10 +370,13 @@ export class GoldSkySubscriber implements BaseSubscriber {
             }
             return json?.data ?? null;
         } catch (err: any) {
-            if (err?.name !== 'AbortError') {
+            if (err?.name !== 'AbortError' && err?.name !== 'TimeoutError') {
                 logger.warn(`GoldSkySubscriber: Fetch failed for ${q.url}`, { error: String(err) });
             }
             return null;
+        } finally {
+            signal.removeEventListener('abort', onAbort);
+            timeoutSignal.removeEventListener('abort', onAbort);
         }
     }
 }


realfishsam:

### What This Does
Adds a 30-second timeout to GoldSky subscriber GraphQL fetch calls using `AbortSignal.timeout()`, combined with the existing caller-provided abort signal via a shared `AbortController`. Prevents indefinite hangs when the GoldSky subgraph endpoint is unresponsive.

### Blast Radius
- **Core**: `core/src/subscriber/external/goldsky.ts` only

### Findings
1. **Signal composition is correct**: The pattern of creating a `combinedController` and wiring both the caller's `signal` and `AbortSignal.timeout(30_000)` to abort it is the standard approach. The `finally` block properly removes both event listeners to avoid leaks.
2. **Node.js compatibility**: `AbortSignal.timeout()` requires Node 17.3+. The project requires Node >=18.0.0, so this is safe. This is the first use of `AbortSignal.timeout()` in the codebase.
3. **TimeoutError handling**: The catch block now also suppresses `TimeoutError` (in addition to `AbortError`), returning `null` silently. This is consistent with the existing behavior of treating fetch failures as non-fatal for polling subscribers. However, a `logger.debug` for timeouts specifically might help debugging slow endpoints in production. Non-blocking.
4. **Hardcoded 30s timeout**: The timeout is not configurable. This matches the existing `DEFAULT_WATCH_TIMEOUT_MS` constant used elsewhere, but a config option would be more flexible. Non-blocking.

### PMXT Pipeline Check
- Field propagation: N/A
- OpenAPI sync: N/A
- Type safety: OK

### Semver Impact
patch -- internal resilience fix, no API or behavioral change

### Risk
Not verified against a live GoldSky endpoint. The 30s timeout is reasonable but may be too aggressive if the subgraph legitimately takes longer under load. Monitor after deploy.

PR #577, opened 2026-05-24
fix: add quoteToken to hyperliquid outcome interface

### LINKED ISSUE #555: Response drift: hyperliquid — outcomes[].quoteToken field present in live API but missing from TypeScript interface
## Exchange
hyperliquid

## Severity
MEDIUM

The `HyperliquidRawOutcome` TypeScript interface does not declare a `quoteToken` field. The normalizer never accesses it, so it is silently ignored during market normalization.

## What The Live API Returns
The `POST https://api.hyperliquid.xyz/info` endpoint with body `{"type":"outcomeMeta"}` returns:
```
outcomes[].outcome: int
outcomes[].name: str
outcomes[].description: str
outcomes[].sideSpecs[].name: str
outcomes[].quoteToken: str      ← present in live response, absent from interface
```

Endpoint tested: `POST https://api.hyperliquid.xyz/info` body `{"type":"outcomeMeta"}`

`quoteToken` is a string (e.g. `"USDC"`) indicating the settlement currency for each prediction-market outcome. The TypeScript interface `HyperliquidRawOutcome` does not include this field, so it is never propagated to `UnifiedMarket`.

## Impact
MEDIUM: Settlement currency information is silently dropped during normalization. For Hyperliquid markets settled in tokens other than the default (if any multi-currency outcomes are introduced), PMXT would have no way to display or act on the currency. Currently causes no functional crash but represents missing data for potential future multi-currency support.

## Suggested Fix
Add `quoteToken` to the `HyperliquidRawOutcome` interface and thread it into the unified market metadata:
```typescript
// In HyperliquidRawOutcome interface
quoteToken: string;

// In normalizer, when building UnifiedMarket
metadata: { ..., settlementCurrency: raw.outcome.quoteToken }
```

---

realfishsam:
Fixes #555

### DIFF
--- a/core/src/exchanges/hyperliquid/fetcher.ts
+++ b/core/src/exchanges/hyperliquid/fetcher.ts
@@ -17,6 +17,7 @@ export interface HyperliquidRawOutcome {
     name: string;           // e.g. "BTC > $100K @ 2026-05-09 06:00 UTC"
     description: string;    // pipe-delimited contract spec
     sideSpecs: HyperliquidRawSideSpec[];
+    quoteToken: string;     // settlement currency, e.g. "USDC"
 }
 
 export interface HyperliquidRawQuestion {


realfishsam:

### What This Does
Adds a `quoteToken: string` field to the `HyperliquidRawOutcome` interface to capture the settlement currency (e.g., "USDC") from the venue API response.

### Blast Radius
- **Core layer only**: `core/src/exchanges/hyperliquid/fetcher.ts` (raw type definition)
- No normalizer, OpenAPI schema, or SDK changes

### Findings
1. The field is added to the raw venue-native interface but there is no evidence it is consumed anywhere downstream (no normalizer reads `quoteToken`, no unified type maps it). This is safe as a type-accuracy fix (making the interface match the real API shape), but it does not surface the data to SDK users.
2. No test coverage for the new field. Acceptable for a raw interface addition, but if the intent (per issue #555) is to expose quote currency to users, additional normalizer + type + SDK work is needed.

### PMXT Pipeline Check
- Field propagation: N/A (raw interface only, not a unified field)
- OpenAPI sync: OK (no BaseExchange.ts change)
- Type safety: OK

### Semver Impact
patch -- additive raw type field, no behavioral change

### Risk
- If the Hyperliquid API does not always include `quoteToken` on every outcome, this should be `quoteToken?: string` (optional). The PR makes it required. Verify against real API responses.

PR #578, opened 2026-05-24
fix: type GeminiRawEvent.series as object|null

### LINKED ISSUE #439: Spec drift: gemini-titan — GeminiRawEvent.series typed as string but live API returns object|null
## Exchange
gemini-titan

## Drift Type
changed_response

## Severity
LOW — Type mismatch causes silent data loss; `series` is always discarded as it can't be treated as a string.

## Their Docs
The Gemini prediction markets API returns `series` on event objects as an **object or null** (not a string). The object contains series-level metadata about the event group.
Source: https://docs.gemini.com/prediction-markets/

## Our Cached Spec / Implementation
`core/src/exchanges/gemini-titan/fetcher.ts` (or the types file) defines `GeminiRawEvent.series` as `series?: string`. Treating an object as a string will produce `[object Object]` or undefined behavior when this field is accessed.

There is no OpenAPI spec for this exchange (see issue #408).

## callApi() References at Risk
None — direct HTTP.

## Hardcoded URL Bypasses
- `GET ${this.baseUrl}/v1/prediction-markets/events` at fetcher.ts:66–68
- `GET ${this.baseUrl}/v1/prediction-markets/events/{eventTicker}` at fetcher.ts:92–94

## Impact
The `series` field is silently mishandled. Any code that stringifies or reads string properties on `event.series` will get garbage. Low severity because the `series` field doesn't currently feed any normalized output field.

---
*Found by automated spec drift audit*

realfishsam:
Fixes #439

### DIFF
--- a/core/src/exchanges/gemini-titan/types.ts
+++ b/core/src/exchanges/gemini-titan/types.ts
@@ -44,7 +44,7 @@ export interface GeminiRawEvent {
     imageUrl?: string;
     type: string;
     category?: string;
-    series?: string;
+    series?: Record<string, any> | null;
     ticker: string;
     status: string;
     resolvedAt?: string | null;


realfishsam:

### What This Does
Changes the type of `GeminiRawEvent.series` from `string` to `Record<string, any> | null` to match the actual Gemini Titan API response shape where `series` is an object (or null), not a string.

### Blast Radius
- **Core layer only**: `core/src/exchanges/gemini-titan/types.ts` (raw type definition)
- No normalizer or downstream code references `series` -- grep confirms it is only defined in this types file and not read by the normalizer

### Findings
No blocking findings. The change correctly widens the type to match observed API behavior.

### PMXT Pipeline Check
- Field propagation: N/A (raw interface only, not mapped to unified type)
- OpenAPI sync: OK (no BaseExchange.ts change)
- Type safety: OK

### Semver Impact
patch -- raw type correction, no behavioral change

### Risk
- If `series` can also be a plain `string` in some API responses (e.g., older events), the type should be `Record<string, any> | string | null` to avoid regressions. Verify against live responses to ensure `string` is truly never returned.

PR #581, opened 2026-05-24
fix: make params optional on compareMarketPrices and fetchRelatedMarkets

### LINKED ISSUE #448: `compareMarketPrices` and `fetchRelatedMarkets` take non-optional `params` while sibling router methods take optional `params`
## Inconsistency

Among the four router methods on `PredictionMarketExchange`, two require a `params` argument and two make it optional. There is no documented reason for the asymmetry.

## Current (PMXT)

`core/src/BaseExchange.ts` lines ~1395–1420:
```typescript
async compareMarketPrices(params: FetchMatchesParams): Promise<PriceComparison[]>       // ← required
async fetchRelatedMarkets(params: FetchMatchesParams): Promise<PriceComparison[]>        // ← required
async fetchMarketMatches(params?: FetchMarketMatchesParams): Promise<MatchResult[]>      // ← optional
async fetchEventMatches(params?: FetchEventMatchesParams): Promise<MatchResult[]>        // ← optional (line ~1425)
```

## Expected (CCXT convention)

CCXT uses a consistent `params={}` default on every method (`ts/src/base/Exchange.ts`). When a method can meaningfully be called with no arguments, the params bag should default to an empty object or be optional. All four router methods can logically be called without configuration (returning all matches / comparisons for the current exchange). They should either all be optional or all required.

## Impact

Callers who write a generic helper that calls all router methods uniformly will hit a TypeScript compile error on the two required-params variants. The inconsistency forces special-casing with no functional justification.

---

realfishsam:
Fixes #448

### DIFF
--- a/core/src/BaseExchange.ts
+++ b/core/src/BaseExchange.ts
@@ -1518,7 +1518,7 @@ export abstract class PredictionMarketExchange {
      * @param params - Match filter parameters (uses relation: 'identity' internally)
      * @returns Array of price comparisons across venues
      */
-    async compareMarketPrices(params: FetchMatchesParams): Promise<PriceComparison[]> {
+    async compareMarketPrices(params?: FetchMatchesParams): Promise<PriceComparison[]> {
         throw new Error("Method compareMarketPrices not implemented.");
     }
 
@@ -1529,7 +1529,7 @@ export abstract class PredictionMarketExchange {
      * @param params - Match filter parameters
      * @returns Array of subset/superset matches with live prices
      */
-    async fetchRelatedMarkets(params: FetchMatchesParams): Promise<PriceComparison[]> {
+    async fetchRelatedMarkets(params?: FetchMatchesParams): Promise<PriceComparison[]> {
         throw new Error("Method fetchRelatedMarkets not implemented.");
     }
 
@@ -1558,7 +1558,7 @@ export abstract class PredictionMarketExchange {
      * @param params - Match filter parameters
      * @returns Array of subset/superset matches with live prices
      */
-    async fetchHedges(params: FetchMatchesParams): Promise<PriceComparison[]> {
+    async fetchHedges(params?: FetchMatchesParams): Promise<PriceComparison[]> {
         throw new Error("Method fetchHedges not implemented.");
     }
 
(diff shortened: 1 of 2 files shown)


realfishsam:
## PR Review: FAIL

### What This Does
Makes the `params` argument optional on `compareMarketPrices`, `fetchRelatedMarkets`, and `fetchHedges` in both `BaseExchange.ts` (abstract declarations) and `Router.ts` (concrete implementations), allowing callers to invoke these methods without arguments for "browse all" mode.

### Blast Radius
- **Core**: `BaseExchange.ts` (3 method signatures), `Router.ts` (3 implementations)
- **OpenAPI spec**: `openapi.yaml` is AST-generated from `BaseExchange.ts` -- signature changes require regeneration
- **SDKs**: Generated TS and Python clients depend on the OpenAPI spec

### Findings

1. **CRITICAL: OpenAPI spec not regenerated.** `BaseExchange.ts` is changed but `openapi.yaml` is not updated. Per CLAUDE.md and the `.github/workflows/openapi-check.yml` CI check, this will fail CI. The PR must run `npm run generate:openapi --workspace=pmxt-core` and include the updated `openapi.yaml` and `method-verbs.json`.

2. **Note on `compareMarketPrices` OpenAPI behavior.** The current OpenAPI spec for `compareMarketPrices` uses a POST with `required: ["args"]` in the request body. Making params optional in TS does not automatically make the `args` array optional in the spec -- the regeneration step should handle this, but it needs verification.

3. **Router.ts uses `params.market` without null-guarding for the `= {}` default.** When `params` defaults to `{}`, `params.market` is `undefined`, so the `if (params.market && !params.marketId)` guard works correctly. No issue here.

4. **Good immutable patterns.** Router.ts already uses `params = { ...params, ... }` spread for reassignment, consistent with project conventions.

### PMXT Pipeline Check
- Field propagation: N/A
- OpenAPI sync: ISSUE (BaseExchange.ts changed, openapi.yaml not regenerated)
- Type safety: OK (optional params with `= {}` defaults are sound)

### Semver Impact
minor -- relaxes required argument to optional, backward-compatible

### Risk
- CI will reject this PR until `openapi.yaml` is regenerated. After regeneration, SDKs should be regenerated too to reflect the optional params.

PR #584, opened 2026-05-24
fix: use camelCase endDateIso in polymarket normalizer

### LINKED ISSUE #557: Response drift: polymarket — end_date_iso fallback uses snake_case but live API field is camelCase endDateIso
## Exchange
polymarket

## Severity
LOW

`utils.ts:99` reads `market.end_date_iso` as the secondary fallback for resolution date when `market.endDate` is absent:
```typescript
resolutionDate: market.endDate
    ? new Date(market.endDate)
    : (market.end_date_iso ? new Date(market.end_date_iso) : new Date()),
```

## What The Live API Returns
The `GET https://gamma-api.polymarket.com/markets?limit=1` response includes:
```
[].endDate: str        ← primary field (present, camelCase)
[].endDateIso: str     ← secondary field (present, camelCase)
```
The fallback field is named `endDateIso` (camelCase) in the live response, not `end_date_iso` (snake_case) as the normalizer reads.

Endpoint tested: `GET https://gamma-api.polymarket.com/markets?limit=1`

## Impact
LOW: The primary `endDate` field is present in the live response, so the fallback is not normally triggered. However, if any market has a null/missing `endDate` (e.g., newly created markets, markets pending settlement, or markets with misconfigured dates), the fallback `market.end_date_iso` evaluates to `undefined` (wrong field name), and the final fallback `new Date()` fires — setting `resolutionDate` to today's date instead of the actual expiry. Markets with null `endDate` will silently display incorrect resolution dates.

## Suggested Fix
Update the fallback to use the correct camelCase field name:
```typescript
resolutionDate: market.endDate
    ? new Date(market.endDate)
    : market.endDateIso
    ? new Date(market.endDateIso)
    : new Date(),
```

---

realfishsam:
Fixes #557

### DIFF
--- a/core/src/exchanges/polymarket/utils.ts
+++ b/core/src/exchanges/polymarket/utils.ts
@@ -96,7 +96,7 @@ export function mapMarketToUnified(event: any, market: any, options: { useQuesti
         description: market.description || event.description,
         slug: typeof market.slug === 'string' && market.slug.length > 0 ? market.slug : undefined,
         outcomes: outcomes,
-        resolutionDate: market.endDate ? new Date(market.endDate) : (market.end_date_iso ? new Date(market.end_date_iso) : new Date()),
+        resolutionDate: market.endDate ? new Date(market.endDate) : (market.endDateIso ? new Date(market.endDateIso) : new Date()),
         volume24h: Number(market.volume24hr || market.volume_24h || 0),
         volume: Number(market.volume || 0),
         liquidity: Number(market.liquidity || market.rewards?.liquidity || 0),


realfishsam:
## PR Review: VERIFIED

### What This Does
Fixes the Polymarket normalizer fallback for resolution date: changes `market.end_date_iso` (snake_case) to `market.endDateIso` (camelCase) to match the actual Gamma API response field name confirmed in the vendor OpenAPI spec (`core/specs/polymarket/PolymarketGammaAPI.yaml` line 1532).

### Blast Radius
- **Polymarket exchange**: `utils.ts` (`mapMarketToUnified`) -- single line change in fallback path
- Only affects markets where `market.endDate` is falsy and `endDateIso` is the only available date field
- No SDK, OpenAPI, or type changes

### Findings
1. **Vendor spec confirms camelCase** -- `PolymarketGammaAPI.yaml` defines `endDateIso` (camelCase), validating this fix.
2. **Raw type interface is stale** -- `PolymarketRawMarket` in `fetcher.ts` (line 55) still defines `end_date_iso?: string` (snake_case). This interface should be updated to `endDateIso` to stay consistent with the vendor spec and this fix. The field is accessed via `[key: string]: unknown` index signature so it works at runtime, but the typed property is now dead code.
3. **No test for this fallback path** -- the `endDateIso` fallback is only hit when `endDate` is missing, which appears to be an edge case. No test was added or exists.

### PMXT Pipeline Check
- Field propagation: N/A (resolutionDate is already in all layers)
- OpenAPI sync: N/A
- Type safety: ISSUE -- `PolymarketRawMarket.end_date_iso` in `fetcher.ts` should be renamed to `endDateIso` to match

### Semver Impact
patch -- bug fix for incorrect field name in fallback path

### Risk
The stale `end_date_iso` property in `PolymarketRawMarket` interface is cosmetic but could mislead future contributors. Low risk since the fix is correct per the vendor spec.

PR #586, opened 2026-05-24
fix: read eventId instead of questionId in myriad normalizer

### LINKED ISSUE #556: Response drift: myriad — normalizer reads raw.questionId but live API field is eventId (always null)
## Exchange
myriad

## Severity
LOW

`normalizer.ts:42` reads `raw.questionId` and uses it as the event linkage field:
```typescript
eventId: raw.questionId ? String(raw.questionId) : undefined,
```

## What The Live API Returns
The `GET https://api-v2.myriadprotocol.com/markets?limit=2` response does **not** include a `questionId` field. The live response contains:
```
data[].eventId: NoneType   ← present but always null in sampled data
```
There is no `questionId` field — the field appears to have been renamed to `eventId`.

Endpoint tested: `GET https://api-v2.myriadprotocol.com/markets?limit=2`

## Impact
LOW: The conditional `raw.questionId ? ...` always evaluates falsy (field is absent), so `UnifiedMarket.eventId` is always `undefined` for Myriad markets. Myriad markets are never grouped under their parent events in PMXT's unified event/market hierarchy. Working but degraded — cross-venue event matching for Myriad markets cannot link to events.

## Suggested Fix
Update the normalizer to read the current field name with a fallback:
```typescript
eventId: raw.questionId
    ? String(raw.questionId)
    : raw.eventId
    ? String(raw.eventId)
    : undefined,
```
Note that `eventId` also appears to be `null` in the live sample — this may indicate that market-to-event linkage is populated only for certain market types on Myriad (e.g., question markets fetched via `/questions/{id}`).

---

realfishsam:
Fixes #556

### DIFF
--- a/core/src/exchanges/myriad/fetcher.ts
+++ b/core/src/exchanges/myriad/fetcher.ts
@@ -17,7 +17,7 @@ export interface MyriadRawMarket {
     volume24h?: number;
     volume?: number;
     liquidity?: number;
-    questionId?: number;
+    eventId?: number;
     topics?: string[];
     outcomes?: MyriadRawOutcome[];
     [key: string]: unknown;
--- a/core/src/exchanges/myriad/normalizer.ts
+++ b/core/src/exchanges/myriad/normalizer.ts
@@ -39,7 +39,7 @@ export class MyriadNormalizer implements IExchangeNormalizer<MyriadRawMarket, My
 
         const um = {
             marketId: `${raw.networkId}:${raw.id}`,
-            eventId: raw.questionId ? String(raw.questionId) : undefined,
+            eventId: raw.eventId ? String(raw.eventId) : undefined,
             title: raw.title || '',
             description: raw.description || '',
             outcomes,


realfishsam:
## PR Review: FAIL

### What This Does
Changes the Myriad normalizer and raw type interface to read `eventId` instead of `questionId` from the API response, fixing the event ID mapping in the normalizer.

### Blast Radius
- **Myriad exchange**: `fetcher.ts` (raw type interface), `normalizer.ts` (normalization logic)
- Affects `eventId` field on all Myriad UnifiedMarket objects

### Findings
1. **BLOCKING: `utils.ts` not updated** -- `core/src/exchanges/myriad/utils.ts` line 68 still reads `market.questionId` to populate `eventId`. This is a parallel normalization path (`mapMarketToUnified` in utils.ts vs `normalizeMarket` in normalizer.ts) that will remain broken after this PR. The same `questionId -> eventId` rename must be applied there.
2. **Raw type update is incomplete** -- the `MyriadRawMarket` interface in `fetcher.ts` changed `questionId` to `eventId`, but the `[key: string]: unknown` index signature means the old field name still compiles. However, code in `utils.ts` still references the typed `questionId` field (which no longer exists on the interface), creating a type inconsistency even though it compiles due to the catch-all index signature.
3. **No test coverage** -- no test was added to verify that `eventId` is correctly populated from the API response.

### PMXT Pipeline Check
- Field propagation: ISSUE -- `utils.ts` is a second normalization path that was not updated
- OpenAPI sync: N/A
- Type safety: ISSUE -- stale `questionId` reference in `utils.ts` after interface rename

### Semver Impact
patch -- bug fix for incorrect field mapping

### Risk
Merging as-is leaves `utils.ts` broken: Myriad markets normalized through the `mapMarketToUnified` utility path will have `eventId: undefined` because `market.questionId` no longer exists on the raw type (it was renamed to `eventId`). This is a data correctness issue.

realfishsam:
Closing — incomplete fix. Missed utils.ts:68. Will resubmit with complete fix.

PR #587, opened 2026-05-24
fix: correct category field mapping in baozi normalizer

### LINKED ISSUE #540: Spec drift: baozi — category field returns tier (official/lab/private) instead of topic category
## Summary

PMXT's Baozi adapter returns a `category` field for markets that maps to the on-chain market tier (`official`, `lab`, `private`), rather than a topic-based category (`crypto`, `sports`, `politics`, etc.). This is a semantic mismatch — the PMXT unified interface expects topic categories, not tier designations.

## Impact

**MEDIUM** — market category filtering and display in PMXT will be incorrect for Baozi markets. Users filtering by "crypto" or "sports" will not see Baozi markets; instead they'll see tier labels that don't match any other exchange's category taxonomy.

## Details

- Field: `category` in the PMXT market object
- Baozi on-chain data: exposes a `tier` (official/lab/private) indicating the market creator type
- Baozi off-chain metadata: the `GET /api/markets/metadata` endpoint (see companion issue) includes a topic category field
- Current PMXT behavior: maps `tier` → `category`, which is semantically wrong
- Correct approach: fetch topic category from off-chain metadata endpoint

## Required Fix

1. Fetch topic category from the Baozi off-chain metadata endpoint (see related issue about the metadata endpoint)
2. Map the topic category to PMXT's `category` field
3. Map `tier` to a separate PMXT field (e.g. `tags` or a Baozi-specific extension field) if the tier information should be preserved

realfishsam:
Fixes #540

### DIFF
--- a/core/src/exchanges/baozi/utils.ts
+++ b/core/src/exchanges/baozi/utils.ts
@@ -408,8 +408,8 @@ export function mapBooleanToUnified(market: BaoziMarket, pubkey: string): Unifie
         volume: totalPoolSol,
         liquidity: totalPoolSol,
         url: `https://baozi.bet/market/${pubkey}`,
-        category: layerName(market.layer),
-        tags: [layerName(market.layer), 'solana', 'pari-mutuel'],
+        category: undefined,
+        tags: [`tier:${layerName(market.layer)}`, 'solana', 'pari-mutuel'],
     };
 
     addBinaryOutcomes(um);
@@ -449,8 +449,8 @@ export function mapRaceToUnified(market: BaoziRaceMarket, pubkey: string): Unifi
         volume: totalPoolSol,
         liquidity: totalPoolSol,
         url: `https://baozi.bet/market/${pubkey}`,
-        category: layerName(market.layer),
-        tags: [layerName(market.layer), 'solana', 'pari-mutuel', 'race'],
+        category: undefined,
+        tags: [`tier:${layerName(market.layer)}`, 'solana', 'pari-mutuel', 'race'],
     };
 
     // For 2-outcome races, add binary convenience getters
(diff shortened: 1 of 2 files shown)


realfishsam:
## PR Review: VERIFIED

### What This Does
Corrects the Baozi normalizer's `category` field mapping: changes `category` from `layerName(market.layer)` (e.g., "official", "lab") to `undefined`, and moves the layer info into tags with a `tier:` prefix (e.g., `tier:official`, `tier:lab`). The rationale is that Baozi's "layer" concept is a trust tier, not a topic category.

### Blast Radius
- **Baozi exchange**: `utils.ts` (`mapBooleanToUnified`, `mapRaceToUnified`) -- both market type mappers updated
- **Tests**: `exchange-normalizers-3.test.ts` -- 6 assertion updates matching the new behavior
- No SDK, OpenAPI, or other exchange changes

### Findings
1. **Tests updated and consistent** -- all 3 test assertions (boolean category, boolean tags, race category) are updated to match the new behavior. Test descriptions are also updated to explain the reasoning.
2. **Semantic improvement** -- using `category` for a trust tier was arguably a misuse of the field. Moving it to `tags` with a `tier:` prefix is cleaner and avoids polluting category-based filtering.
3. **Breaking change for consumers** -- any downstream code filtering Baozi markets by `category === 'official'` or `category === 'lab'` will break. Since `category` becomes `undefined`, and tags change from `['official', ...]` to `['tier:official', ...]`, both access patterns are affected.
4. **No migration path** -- there is no deprecation warning or documentation of this change for SDK consumers.

### PMXT Pipeline Check
- Field propagation: OK -- `category` and `tags` are already in all 3 pipeline layers
- OpenAPI sync: N/A
- Type safety: OK

### Semver Impact
minor -- changes the semantics of `category` and `tags` for Baozi markets; consumers relying on the old values will need to update

### Risk
Consumers filtering by `category` or tag values for Baozi markets will silently get different results. No verification that SDK consumers are unaffected.

PR #588, opened 2026-05-24
fix: read liquidity_dollars instead of deprecated liquidity in kalshi

### LINKED ISSUE #554: Response drift: kalshi — market.liquidity field renamed to liquidity_dollars, liquidity always 0
## Exchange
kalshi

## Severity
LOW

`normalizer.ts:88` accesses `market.liquidity` with a `|| 0` fallback:
```typescript
liquidity: Number(market.liquidity || 0),
```

## What The Live API Returns
The `GET https://api.elections.kalshi.com/trade-api/v2/markets?limit=1` response does **not** include a `liquidity` field. The live response contains:
```
markets[].liquidity_dollars: str   ← present (dollar-denominated string)
```
`market.liquidity` is absent — the field was renamed to `liquidity_dollars`.

Endpoint tested: `GET https://api.elections.kalshi.com/trade-api/v2/markets?limit=1`

## Impact
LOW: The `|| 0` fallback fires on every market normalization. `UnifiedMarket.liquidity` is always `0` for all Kalshi markets. Any feature that displays or filters by liquidity (market ordering, liquidity-gated UI) will show Kalshi as having zero liquidity.

## Suggested Fix
Update the normalizer to read the current field name, with a parse step since the value is a dollar string:
```typescript
liquidity: parseFloat(market.liquidity_dollars || market.liquidity || '0') || 0,
```
This handles both the new field name and retains backwards compatibility if the old field ever reappears.

---

realfishsam:
Fixes #554

### DIFF
--- a/core/src/exchanges/kalshi/normalizer.ts
+++ b/core/src/exchanges/kalshi/normalizer.ts
@@ -86,7 +86,7 @@ export class KalshiNormalizer implements IExchangeNormalizer<KalshiRawEvent, Kal
             resolutionDate: new Date(market.expiration_time),
             volume24h: parseFloat(market.volume_24h_fp ?? '') || Number(market.volume_24h || market.volume || 0),
             volume: parseFloat(market.volume_fp ?? '') || Number(market.volume || 0),
-            liquidity: Number(market.liquidity || 0),
+            liquidity: parseFloat(String(market.liquidity_dollars || market.liquidity || '0')) || 0,
             openInterest: parseFloat(market.open_interest_fp ?? '') || Number(market.open_interest || 0),
             url: `https://kalshi.com/events/${event.event_ticker}`,
             category: event.category,
@@ -361,7 +361,7 @@ function eventVolume(event: any): number {
 }
 
 function eventLiquidity(event: any): number {
-    return (event.markets || []).reduce((sum: number, m: any) => sum + (parseFloat(m.open_interest_fp ?? '') || Number(m.open_interest || m.liquidity || 0)), 0);
+    return (event.markets || []).reduce((sum: number, m: any) => sum + (parseFloat(m.open_interest_fp ?? '') || parseFloat(m.liquidity_dollars || m.open_interest || m.liquidity || '0') || 0), 0);
 }
 
 function eventNewest(event: any): number {


realfishsam:

### What This Does
Updates the Kalshi normalizer to prefer `liquidity_dollars` over the deprecated `liquidity` field for both per-market liquidity and the event-level liquidity aggregation function.

### Blast Radius
- **Kalshi exchange**: `normalizer.ts` -- two changes:
  1. `normalizeRawMarket`: liquidity line now reads `liquidity_dollars` first, falls back to `liquidity`
  2. `eventLiquidity`: aggregation function updated with same precedence
- No SDK, OpenAPI, or other exchange changes

### Findings
1. **Vendor spec confirms both fields are deprecated** -- per `core/specs/kalshi/Kalshi.yaml` (line 6736-6740), `liquidity_dollars` is marked `deprecated` and "will always return '0.0000'". The old `liquidity` field (line 6734-6735) is also deprecated and "will always return 0". Both fields are deprecated and return zero values according to the spec.
2. **Type safety concern** -- `KalshiRawMarket` in `fetcher.ts` (line 31) only defines `liquidity?: number`. There is no `liquidity_dollars` property. The field is accessed via the `[key: string]: unknown` index signature, which means there is no compile-time guarantee it exists or has the expected type. Adding `liquidity_dollars?: string` to the interface would be cleaner.
3. **parseFloat on potentially undefined** -- in `normalizeRawMarket`, the expression `parseFloat(String(market.liquidity_dollars || market.liquidity || '0'))` will work but is convoluted. `String(undefined)` produces `"undefined"`, which `parseFloat` returns `NaN` for, then `|| 0` catches it. It works but is fragile.
4. **eventLiquidity has a different pattern** -- uses `parseFloat(m.liquidity_dollars || m.open_interest || m.liquidity || '0')` which passes potentially non-string values directly to `parseFloat`. Since `m.open_interest` and `m.liquidity` are typed as `number | undefined`, passing a number to `parseFloat` is fine (it gets stringified), but the mixing of types in the fallback chain is not clean.
5. **No tests added** -- no test coverage for the liquidity field change.

### PMXT Pipeline Check
- Field propagation: N/A (liquidity is already in all layers)
- OpenAPI sync: N/A
- Type safety: ISSUE -- `liquidity_dollars` not in `KalshiRawMarket` interface

### Semver Impact
patch -- internal field preference change; no public API impact unless liquidity values were previously non-zero and are now zero (or vice versa)

### Risk
Both `liquidity` and `liquidity_dollars` are documented as deprecated and returning zero in the Kalshi vendor spec. This change may be moot if Kalshi has already zeroed both fields. Cannot verify without a live API call whether `liquidity_dollars` actually returns meaningful data despite the spec claiming it returns "0.0000".

PR #592, opened 2026-05-24
fix: handle missing image_url in kalshi event normalizer

### LINKED ISSUE #442: Response drift: kalshi — event.image_url absent from live /events response, no fallback
## Exchange
kalshi

## Severity
MEDIUM

`normalizer.ts:115` accesses `raw.image_url` with no fallback:
```typescript
image: raw.image_url,
```

## What The Live API Returns
The `GET https://api.elections.kalshi.com/trade-api/v2/events?limit=1` response does **not** include an `image_url` field.

Endpoint tested: `GET https://api.elections.kalshi.com/trade-api/v2/events?limit=1`

Fields present on event objects:
```
events[].available_on_brokers: bool
events[].category: str
events[].collateral_return_type: str
events[].event_ticker: str
events[].last_updated_ts: str
events[].mutually_exclusive: bool
events[].series_ticker: str
events[].strike_period: str
events[].sub_title: str
events[].title: str
```

`image_url` is not present.

## Impact
MEDIUM: All kalshi events will have `image: undefined`. Downstream UI components that render event images receive `undefined` instead of a URL string, which may show broken image placeholders or throw if the field is expected to be a string.

## Suggested Fix
Add a fallback: `image: raw.image_url ?? undefined` is already effectively what happens, but to make it explicit and intentional — or check if the Kalshi `/markets` endpoint (which may include `image_url` on individual markets) can supply this and thread it through event normalization.

---

realfishsam:
Fixes #442

### DIFF
--- a/core/src/exchanges/kalshi/normalizer.ts
+++ b/core/src/exchanges/kalshi/normalizer.ts
@@ -114,7 +114,7 @@ export class KalshiNormalizer implements IExchangeNormalizer<KalshiRawEvent, Kal
                 ? markets.reduce((sum, m) => sum + (m.volume ?? 0), 0)
                 : undefined,
             url: `https://kalshi.com/events/${raw.event_ticker}`,
-            image: raw.image_url,
+            image: raw.image_url ?? undefined,
             category: raw.category,
             tags: raw.tags || [],
         };


realfishsam:
## PR Review: VERIFIED

### What This Does
Fixes the Kalshi event normalizer to handle a missing `image_url` field by coercing `null` to `undefined` via the nullish coalescing operator. Previously, a `null` value from the API would propagate as `null` into `UnifiedEvent.image`, which is typed as `string | undefined` (not `string | null`).

### Blast Radius
- `core/src/exchanges/kalshi/normalizer.ts` -- single line change in `normalizeEvent`

### Findings
No blocking findings. The change is correct: `raw.image_url ?? undefined` converts `null` to `undefined` while passing through valid strings unchanged. This aligns with the `image?: string` type on `UnifiedEvent`.

### PMXT Pipeline Check
- Field propagation: N/A (no new fields)
- OpenAPI sync: N/A
- Type safety: OK (fixes a null-vs-undefined type mismatch)

### Semver Impact
patch -- bug fix for null handling.

### Risk
None. Single-line, narrowly scoped fix.

PR #598, opened 2026-05-24
fix: remove dangerous defaults from Python create_order/build_order

### LINKED ISSUE #466: SDK drift: Python `create_order`/`build_order` have dangerous defaults (`side="buy"`, `amount=0`); TypeScript requires them
## Drift

Python `create_order` and `build_order` supply default values for `side`, `type`, and `amount` — required, high-stakes order parameters. TypeScript's `CreateOrderParams` declares these fields as required with no defaults, forcing callers to be explicit.

## TypeScript SDK

`sdks/typescript/pmxt/models.ts`, lines 470–491 — `CreateOrderParams` interface with required fields:

```typescript
// sdks/typescript/pmxt/models.ts lines 470–491
export interface CreateOrderParams {
    marketId?: string;
    outcomeId?: string;
    side: 'buy' | 'sell';         // required — no default
    type: 'market' | 'limit';     // required — no default
    amount: number;                // required — no default
    price?: number;
    fee?: number;
    outcome?: MarketOutcome;
}
```

`sdks/typescript/pmxt/client.ts`, line 1435:
```typescript
async createOrder(params: any): Promise<Order>
async buildOrder(params: CreateOrderParams & { outcome?: MarketOutcome }): Promise<BuiltOrder>
```

## Python SDK

`sdks/python/pmxt/client.py`, lines 2300–2310 — `create_order` with defaults:

```python
# sdks/python/pmxt/client.py lines 2300–2310
async def create_order(
    self,
    market_id: Optional[str] = None,
    outcome_id: Optional[str] = None,
    side: str = "buy",          # dangerous default
    type: str = "market",       # dangerous default
    amount: float = 0,          # dangerous default (zero-size order)
    price: Optional[float] = None,
    fee: Optional[float] = None,
    outcome: Optional[MarketOutcome] = None,
    **kwargs,
) -> Order:
```

`build_order` has the same signature and same defaults (client.py lines ~2330–2340).

## Expected

`side`, `type`, and `amount` should be required (no default) in Python just as they are in TypeScript. Callers must explicitly supply them. If a default is kept for backward-compatibility reasons, at minimum `amount=0` should raise a `ValidationError` before sending to the sidecar.

## Impact

A caller who forgets `side` silently places a buy order. A caller who forgets `amount` silently submits a zero-size order to the exchange. Both behaviours are financially dangerous and differ from TypeScript, where the compiler enforces that these fields are supplied.

realfishsam:
Fixes #466


realfishsam:

### What This Does
Removes dangerous default values from `create_order` and `build_order` in the Python SDK. Previously, `side` defaulted to `"buy"`, `type` defaulted to `"market"`, and `amount` defaulted to `0`. These defaults were hazardous -- a user who forgot to specify `side` or `type` would silently execute a market buy order. Now all three parameters are keyword-only (enforced by the bare `*` separator) with no defaults, requiring explicit values.

### Blast Radius
- `sdks/python/pmxt/client.py` -- `create_order()` and `build_order()` signatures

### Findings
1. **Breaking change: all three params (`side`, `type`, `amount`) are now required keyword-only.** Existing code calling `exchange.create_order(market_id, outcome_id, "buy", "limit", 10, 0.55)` with positional args will break because the bare `*` forces everything after `outcome_id` to be keyword-only. This is intentional and correct for safety.
2. **Merge conflict with PR 600.** PR 600 renames `type` to `order_type` in the same functions. If PR 598 merges first, PR 600 will need rebasing. If PR 600 merges first, this PR will need rebasing. Both PRs modify the same lines. The desired end state should combine both changes: keyword-only `order_type` (not `type`) with no defaults.

### PMXT Pipeline Check
- Field propagation: N/A
- OpenAPI sync: N/A
- Type safety: OK -- stronger parameter requirements

### Semver Impact
major -- removes default values and adds keyword-only constraint; breaks callers relying on positional args or omitted parameters

### Risk
Merge conflict with PR 600 is certain. Coordinate merge order. No tests are added in this PR to verify the new required-keyword behavior.

PR #600, opened 2026-05-24
fix: type safety and naming fixes in Python client.py

### LINKED ISSUE #449: Python `create_order` parameter named `type` shadows Python built-in
## Inconsistency

The Python SDK's `create_order` method uses `type` as a parameter name. `type` is a Python built-in function; using it as a local name shadows the built-in inside the method body and is a well-known Python anti-pattern flagged by linters.

## Current (PMXT)

`sdks/python/pmxt/client.py` line 2301 (approximate):
```python
def create_order(
    self,
    market_id: str,
    outcome_id: str,
    side: Literal["buy", "sell"],
    type: Literal["market", "limit"] = "market",   # ← shadows built-in
    amount: float = None,
    price: float = None,
    params: dict = None,
) -> Order:
```

## Expected (CCXT convention)

CCXT's Python implementation uses `order_type` or positional ordering that avoids the conflict. The idiomatic Python name for this parameter is `order_type`. CCXT's `create_order` signature (`Exchange.py`):
```python
def create_order(self, symbol: str, type: str, side: str, amount: float, price: float = None, params={})
```

Note: CCXT also uses `type` here, making this a shared CCXT issue; however CCXT wraps the parameter with `order_type` in many helper methods. Regardless, PMXT should adopt `order_type` to avoid the shadowing issue in its own codebase.

## Impact

Tools such as `mypy`, `pylint`, and `flake8` will warn on this. Any code inside `create_order` that calls the built-in `type()` function will silently call the parameter instead, introducing a subtle runtime bug risk.

---

### LINKED ISSUE #452: Python `watch_address` uses bare `list[str]` type hint and `= None` default — should use `Optional[List[str]]`
## Inconsistency

`watch_address` in the Python SDK uses the bare `list[str]` generic (Python 3.9+ only) with a `= None` default. This is both a compatibility issue (breaks Python 3.8) and a type-correctness issue (`list[str]` does not include `None`, so the annotation is wrong for the given default).

## Current (PMXT)

`sdks/python/pmxt/client.py` line ~2030:
```python
def watch_address(self, address: str, types: list[str] = None):
```

Two problems:
1. `list[str]` as a generic requires Python ≥ 3.9; the rest of the SDK imports `List` from `typing` for 3.8 compatibility.
2. The annotation `list[str]` does not permit `None`, but the default is `None`. The correct annotation is `Optional[List[str]]`.

## Expected (CCXT convention)

CCXT Python uses `typing.List` and `typing.Optional` throughout for 3.8 compatibility. Consistent with the rest of `sdks/python/pmxt/client.py` and `sdks/python/pmxt/models.py`:
```python
from typing import List, Optional

def watch_address(self, address: str, types: Optional[List[str]] = None):
```

## Impact

Code using this method will fail to import or raise a `TypeError` on Python 3.8. Static type checkers (`mypy`) will flag the `None` default as incompatible with the `list[str]` annotation.

---

### LINKED ISSUE #456: Python `MarketFetchParams` / `EventFetchParams` use snake_case keys but server expects camelCase
## Inconsistency

`MarketFetchParams` and `EventFetchParams` in the Python SDK define TypedDict keys in snake_case (`market_id`, `outcome_id`, `event_id`, `search_in`). If these dicts are passed to the HTTP server as-is, the server will receive snake_case keys where it expects camelCase (`marketId`, `outcomeId`, `eventId`, `searchIn`).

## Current (PMXT)

`sdks/python/pmxt/models.py` lines 555–584:
```python
class MarketFetchParams(TypedDict, total=False):
    query: str
    limit: int
    offset: int
    sort: Literal["volume", "liquidity", "newest"]
    status: Literal["active", "inactive", "closed", "all"]
    search_in: Literal["title", "description", "both"]  # ← snake_case
    slug: str
    market_id: str    # ← snake_case
    outcome_id: str   # ← snake_case
    event_id: str     # ← snake_case
    category: str
    tags: List[str]
    filter: MarketFilterCriteria

class EventFetchParams(TypedDict, total=False):
    query: str
    limit: int
    ...
    event_id: str     # ← snake_case
    search_in: Literal["title", "description", "both"]  # ← snake_case
```

The TypeScript SDK uses camelCase in its equivalent types (`sdks/typescript/pmxt/models.ts`), consistent with what the server expects.

## Expected (CCXT convention)

CCXT's Python SDK translates parameter names to the correct casing before sending them. Python-idiomatic TypedDict keys are acceptable for the public API, but the client must convert `market_id → marketId`, `search_in → searchIn`, etc. before the HTTP request. Either the TypedDict keys should be camelCase (matching the wire format), or the client layer must perform the conversion explicitly.

## Impact

Calls to `fetchMarkets(market_id="abc")` will silently send `market_id` to the server, which ignores it (expects `marketId`). The market-ID filter is silently a no-op. This is a data-loss bug, not merely a style issue.

---

### LINKED ISSUE #496: SDK drift: `OrderBook` datetime field named `datetime` in TypeScript but `dt` in Python
## Drift

The ISO 8601 datetime field of `OrderBook` has different names across the two SDKs. TypeScript calls it `datetime`; Python calls it `dt`. This is not a camelCase→snake_case conversion — `datetime` in snake_case is still `datetime`, not `dt`. Users who read `orderBook.datetime` in TypeScript and try `order_book.dt` in Python (or vice versa) will find the field under a completely different name.

## TypeScript SDK

`sdks/typescript/pmxt/models.ts`, line 155:

```typescript
export interface OrderBook {
    bids: OrderLevel[];
    asks: OrderLevel[];
    timestamp?: number;
    /** ISO 8601 datetime string of the snapshot (CCXT-compatible) */
    datetime?: string;
}
```

## Python SDK

`sdks/python/pmxt/models.py`, lines 264–278:

```python
@dataclass
class OrderBook:
    bids: List[OrderLevel]
    asks: List[OrderLevel]
    timestamp: Optional[int] = None
    dt: Optional[str] = None
    """ISO 8601 datetime string (CCXT-compatible)"""
```

The `_SNAKE_TO_CAMEL` mapping in `sdks/python/pmxt/client.py`, line 65, maps `dt → datetime` when serialising to JSON for the sidecar — but the public Python attribute exposed to users is `dt`, not `datetime`.

## Expected

Both SDKs should expose the field under the same normalised name. After camelCase→snake_case conversion, `datetime` stays `datetime`. Python should name the field `datetime` (or at minimum expose a `datetime` alias), not `dt`.

## Impact

Any user code that accesses `order_book.datetime` in Python will get `AttributeError`; they must know to use `order_book.dt` instead. Code that works on both SDKs cannot use a common field name for this property.

---

### LINKED ISSUE #563: [python-types] client.py: 9 router proxy methods return List[Any] and 3 watch methods have untyped callback
## File
`sdks/python/pmxt/client.py`

> **Note:** This is an addendum to #227, which covers `params: Optional[dict]` and other issues in the same file. These findings were not included there.

## Excessive `Any` — Router Proxy Methods Return `List[Any]` (CRITICAL)

The `Exchange` base class exposes 9 router-related methods that return `List[Any]`. The `Router` subclass properly overrides these with concrete types, but the base class is the declared type for every exchange instance. Users who call these methods on `Polymarket`, `Kalshi`, etc. (which do not override them) get `List[Any]` — no IDE completion on results.

| Line | Method | Current return | Should be |
|------|--------|---------------|-----------|
| 1135 | `fetch_market_matches` | `List[Any]` | `List[MatchResult]` |
| 1158 | `fetch_matches` | `List[Any]` | `List[MatchResult]` |
| 1180 | `fetch_event_matches` | `List[Any]` | `List[EventMatchResult]` |
| 1203 | `compare_market_prices` | `List[Any]` | `List[PriceComparison]` |
| 1225 | `fetch_related_markets` | `List[Any]` | `List[MatchResult]` |
| 1247 | `fetch_matched_markets` | `List[Any]` | `List[MatchResult]` |
| 1270 | `fetch_matched_prices` | `List[Any]` | `List[PriceComparison]` |
| 1293 | `fetch_hedges` | `List[Any]` | `List[PriceComparison]` |
| 1315 | `fetch_arbitrage` | `List[Any]` | `List[ArbitrageOpportunity]` |

All model classes (`MatchResult`, `EventMatchResult`, `PriceComparison`, `ArbitrageOpportunity`) are already imported from `.models` at the top of `client.py`.

## Untyped `callback` Parameter — Watch Methods (HIGH)

Three watch methods accept `callback: Optional[Any]`. The `callback` parameter is not documented as to what signature it should have, making it impossible for users to know what function to pass, and for type checkers to validate call sites.

| Line | Method | Parameter issue |
|------|--------|----------------|
| 2143 | `watch_prices(self, market_address: str, callback: Optional[Any] = None) -> Any` | `callback` untyped; return type `Any` |
| 2179 | `watch_user_positions(self, callback: Optional[Any] = None) -> List[Position]` | `callback` untyped |
| 2216 | `watch_user_transactions(self, callback: Optional[Any] = None) -> Any` | `callback` untyped; return type `Any` |

## Impact

- Every caller of the 9 router proxy methods on any non-Router exchange gets `Any` back — zero IDE autocompletion on results.
- `watch_prices` and `watch_user_transactions` return `Any`, leaving the full shape of streaming payloads invisible to type checkers.
- `callback` being `Optional[Any]` means IDEs cannot validate whether a user-supplied callback has the right signature.

### Router proxy methods
```python
from .models import MatchResult, EventMatchResult, PriceComparison, ArbitrageOpportunity  # already imported

def fetch_market_matches(self, params: Optional[Dict[str, Any]] = None, **kwargs) -> List[MatchResult]: ...
def fetch_event_matches(self, params: Optional[Dict[str, Any]] = None, **kwargs) -> List[EventMatchResult]: ...
def compare_market_prices(self, params: Dict[str, Any], **kwargs) -> List[PriceComparison]: ...
def fetch_hedges(self, params: Dict[str, Any], **kwargs) -> List[PriceComparison]: ...
def fetch_arbitrage(self, params: Optional[Dict[str, Any]] = None, **kwargs) -> List[ArbitrageOpportunity]: ...
# etc.
```

### Watch methods
```python
from typing import Callable

def watch_prices(
    self,
    market_address: str,
    callback: Optional[Callable[[Dict[str, Any]], None]] = None,
) -> Dict[str, Any]: ...

def watch_user_transactions(
    self,
    callback: Optional[Callable[[Dict[str, Any]], None]] = None,
) -> Dict[str, Any]: ...
```

---

realfishsam:
Fixes #563
Fixes #449
Fixes #452
Fixes #456
Fixes #496

### DIFF
--- a/sdks/python/pmxt/client.py
+++ b/sdks/python/pmxt/client.py
@@ -12,7 +12,7 @@ import time
 import urllib.error
 from abc import ABC
 from datetime import datetime
-from typing import List, Optional, Dict, Any, Literal, Union
+from typing import Callable, List, Optional, Dict, Any, Literal, Union
 
 # Add generated client to path
 _GENERATED_PATH = os.path.join(os.path.dirname(__file__), "..", "generated")
@@ -46,6 +46,10 @@ from .models import (
     EventFilterFunction,
     SubscribedAddressSnapshot,
     FirehoseEvent,
+    MatchResult,
+    EventMatchResult,
+    PriceComparison,
+    ArbitrageOpportunity,
 )
 from .constants import LOCAL_URL, resolve_pmxt_base_url
 from .errors import PmxtError, from_server_error
@@ -62,7 +66,6 @@ _SNAKE_TO_CAMEL = {
     'price_change_24h': 'priceChange24h',
     'unrealized_pnl': 'unrealizedPnL',
     'realized_pnl': 'realizedPnL',
-    'dt': 'datetime',
 }
 
 
@@ -72,6 +75,15 @@ def _snake_to_camel(name: str) -> str:
         return _SNAKE_TO_CAMEL[name]
     parts = name.split('_')
     return parts[0] + ''.join(p.title() for p in parts[1:])
+
+
+def _convert_params_to_camel(params: Dict[str, Any]) -> Dict[str, Any]:
+    """Convert snake_case param keys to camelCase for the sidecar wire format.
+
+    Keys that are already camelCase or single-word pass through unchanged.
+    Nested dicts/lists are left as-is -- only top-level keys are converted.
+    """
+    return {_snake_to_camel(k): v for k, v in params.items()}
 
 
 def _auto_convert(cls, raw: Dict[str, Any], **overrides):
@@ -739,7 +751,7 @@ class Exchange(ABC):
             if kwargs:
                 params = {**(params or {}), **kwargs}
             if params is not None:
-                args.append(params)
+                args.append(_convert_params_to_camel(params))
             body: dict = {"args": args}
             creds = self._get_credentials_dict()
             if creds:
@@ -762,7 +774,7 @@ class Exchange(ABC):
             if kwargs:
                 params = {**(params or {}), **kwargs}
             if params is not None:
-                args.append(params)
+                args.append(_convert_params_to_camel(params))
             body: dict = {"args": args}
             creds = self._get_credentials_dict()
             if creds:
@@ -789,7 +801,7 @@ class Exchange(ABC):
             if kwargs:
                 params = {**(params or {}), **kwargs}
             if params is not None:
-                args.append(params)
+                args.append(_convert_params_to_camel(params))
             body: dict = {"args": args}
             creds = self._get_credentials_dict()
             if creds:
@@ -812,7 +824,7 @@ class Exchange(ABC):
             if kwargs:
                 params = {**(params or {}), **kwargs}
             if params is not None:
-                args.append(params)
+                args.append(_convert_params_to_camel(params))
             body: dict = {"args": args}
             creds = self._get_credentials_dict()
             if creds:
@@ -835,7 +847,7 @@ class Exchange(ABC):
             if kwargs:
                 params = {**(params or {}), **kwargs}
             if params is not None:
-                args.append(params)
+                args.append(_convert_params_to_camel(params))
             body: dict = {"args": args}
             creds = self._get_credentials_dict()
             if creds:
@@ -1132,7 +1144,7 @@ class Exchange(ABC):
         except ApiException as e:
             raise self._parse_api_exception(e) from None
 
-    def fetch_market_matches(self, params: Optional[dict] = None, **kwargs) -> List[Any]:
+    def fetch_market_matches(self, params: Optional[dict] = None, **kwargs) -> List[MatchResult]:
         try:
             args = []
             if kwargs:
@@ -1155,7 +1167,7 @@ class Exchange(ABC):
         except ApiException as e:
             raise self._parse_api_exception(e) from None
 
-    def fetch_matches(self, params: dict, **kwargs) -> List[Any]:
+    def fetch_matches(self, params: dict, **kwargs) -> List[MatchResult]:
         try:
             args = []
             if kwargs:
@@ -1177,7 +1189,7 @@ class Exchange(ABC):
         except ApiException as e:
             raise self._parse_api_exception(e) from None
 
-    def fetch_event_matches(self, params: Optional[dict] = None, **kwargs) -> List[Any]:
+    def fetch_event_matches(self, params: Optional[dict] = None, **kwargs) -> List[EventMatchResult]:
         try:
             args = []
             if kwargs:
@@ -1200,7 +1212,7 @@ class Exchange(ABC):
         except ApiException as e:
             raise self._parse_api_exception(e) from None
 
-    def compare_market_prices(self, params: dict, **kwargs) -> List[Any]:
+    def compare_market_prices(self, params: dict, **kwargs) -> List[PriceComparison]:
         try:
             args = []
             if kwargs:
@@ -1222,7 +1234,7 @@ class Exchange(ABC):
         except ApiException as e:
             raise self._parse_api_exception(e) from None
 
-    def fetch_related_markets(self, params: dict, **kwargs) -> List[Any]:
+    def fetch_related_markets(self, params: dict, **kwargs) -> List[MatchResult]:
         try:
             args = []
             if kwargs:
@@ -1244,7 +1256,7 @@ class Exchange(ABC):
         except ApiException as e:
             raise self._parse_api_exception(e) from None
 
-    def fetch_matched_markets(self, params: Optional[dict] = None, **kwargs) -> List[Any]:
+    def fetch_matched_markets(self, params: Optional[dict] = None, **kwargs) -> List[MatchResult]:
         try:
             args = []
             if kwargs:
@@ -1267,7 +1279,7 @@ class Exchange(ABC):
         except ApiException as e:
             raise self._parse_api_exception(e) from None
 
-    def fetch_matched_prices(self, params: Optional[dict] = None, **kwargs) -> List[Any]:
+    def fetch_matched_prices(self, params: Optional[dict] = None, **kwargs) -> List[PriceComparison]:
         try:
             args = []
             if kwargs:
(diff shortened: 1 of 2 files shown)


realfishsam:

### What This Does
Multiple type safety and naming fixes in the Python client: (1) renames `type` param to `order_type` in `create_order`/`build_order` to avoid shadowing the Python builtin, (2) renames `OrderBook.dt` field to `OrderBook.datetime` in models.py and removes the `'dt': 'datetime'` mapping from `_SNAKE_TO_CAMEL`, (3) adds `_convert_params_to_camel()` helper and applies it to 5 fetch methods (fetchMarkets, fetchMarketsPaginated, fetchEvents, fetchMarket, fetchEvent) so snake_case kwargs are converted to camelCase for the sidecar wire format, (4) adds concrete return types (MatchResult, PriceComparison, etc.) to 8 router methods that previously returned `List[Any]`, (5) types `watch_*` callback parameters with `Callable` and return types with `Dict[str, Any]`, (6) fixes `list[str]` to `Optional[List[str]]` on `watch_address`.

### Blast Radius
- `sdks/python/pmxt/client.py` -- parameter renames, param conversion, return types
- `sdks/python/pmxt/models.py` -- `OrderBook.dt` renamed to `OrderBook.datetime`

### Findings
1. **Breaking change: `type` renamed to `order_type`.** All callers using `exchange.create_order(type="limit")` must update to `exchange.create_order(order_type="limit")`. This is the right move to avoid shadowing Python's `type` builtin, but it's a breaking change.
2. **Breaking change: `OrderBook.dt` renamed to `OrderBook.datetime`.** Any code accessing `orderbook.dt` will break. However, `'dt': 'datetime'` is removed from `_SNAKE_TO_CAMEL`, which means `_auto_convert` will now try to map the `datetime` field from the camelCase key `datetime` (which is identity). This is correct because the sidecar sends `"datetime"` as the JSON key. The old mapping was `dt -> datetime` in the mapping table so that the snake_case field `dt` would look up the camelCase key `datetime` in the raw dict. Now that the field IS named `datetime`, the identity mapping works.
3. **Merge conflict with PR 598.** PR 598 makes `side`, `type`, `amount` keyword-only with no defaults. PR 600 renames `type` to `order_type` but keeps the defaults (`"buy"`, `"market"`, `0`). The intended end state should be keyword-only `order_type` with no defaults. These PRs must be merged carefully.
4. **`_convert_params_to_camel` only converts top-level keys.** This is documented in the docstring and consistent with the sidecar's expectations. Nested filter objects (e.g., `filter={"text": "..."}`) pass through as-is, which is correct because nested keys are already camelCase in the filter schema.
5. **New type imports (`MatchResult`, `EventMatchResult`, `PriceComparison`, `ArbitrageOpportunity`) from models.py.** Verified these types exist in the Python models.py already. Import is correct.
6. **`watch_address` types param changed from `list[str] = None` to `Optional[List[str]] = None`.** This is a Python 3.8 compatibility fix (bare `list[str]` requires 3.9+). Good fix.

### PMXT Pipeline Check
- Field propagation: N/A -- no new unified fields
- OpenAPI sync: N/A -- SDK-only changes
- Type safety: OK -- replaces `Any` with concrete types throughout

### Semver Impact
major -- `type` to `order_type` rename and `dt` to `datetime` rename are both breaking changes to the public API

### Risk
Merge conflict with PR 598 is certain (both touch `create_order`/`build_order` signatures). The `_convert_params_to_camel` helper is only applied to 5 methods -- other methods that accept `params` dicts (e.g., `fetch_order_book`, `fetch_trades`, `fetch_ohlcv`) still pass params as-is. Verify those methods don't need the same treatment, or document why they're excluded. No tests added for the camelCase conversion.

PR #602, opened 2026-05-24
fix: read eventId instead of questionId in myriad normalizer

### LINKED ISSUE #556: Response drift: myriad — normalizer reads raw.questionId but live API field is eventId (always null)

realfishsam:
Fixes #556

### DIFF
--- a/core/src/exchanges/myriad/fetcher.ts
+++ b/core/src/exchanges/myriad/fetcher.ts
@@ -17,7 +17,7 @@ export interface MyriadRawMarket {
     volume24h?: number;
     volume?: number;
     liquidity?: number;
-    questionId?: number;
+    eventId?: number;
     topics?: string[];
     outcomes?: MyriadRawOutcome[];
     [key: string]: unknown;
--- a/core/src/exchanges/myriad/normalizer.ts
+++ b/core/src/exchanges/myriad/normalizer.ts
@@ -39,7 +39,7 @@ export class MyriadNormalizer implements IExchangeNormalizer<MyriadRawMarket, My
 
         const um = {
             marketId: `${raw.networkId}:${raw.id}`,
-            eventId: raw.questionId ? String(raw.questionId) : undefined,
+            eventId: raw.eventId ? String(raw.eventId) : undefined,
             title: raw.title || '',
             description: raw.description || '',
             outcomes,
--- a/core/src/exchanges/myriad/utils.ts
+++ b/core/src/exchanges/myriad/utils.ts
@@ -65,7 +65,7 @@ export function mapMarketToUnified(market: any): UnifiedMarket | null {
 
     const um = {
         marketId: `${market.networkId}:${market.id}`,
-        eventId: market.questionId ? String(market.questionId) : undefined,
+        eventId: market.eventId ? String(market.eventId) : undefined,
         title: market.title || '',
         description: market.description || '',
         outcomes,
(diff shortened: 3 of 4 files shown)


realfishsam:
## PR Review: FAIL

### What This Does
Renames `questionId` to `eventId` across three Myriad files to match the current live API response shape. This is v2 of PR #586, which missed `utils.ts:68`.

The three changes:
1. `fetcher.ts:20` -- `MyriadRawMarket` interface field renamed from `questionId` to `eventId`
2. `normalizer.ts:42` -- `normalizeMarket()` now reads `raw.eventId` instead of `raw.questionId`
3. `utils.ts:68` -- `mapMarketToUnified()` now reads `market.eventId` instead of `market.questionId` (the line v1 missed)

### Blast Radius
- Contained to `core/src/exchanges/myriad/` -- three files, one line each.
- No other exchanges are affected. The `questionId` references in `metaculus/`, `hyperliquid/`, and `opinion/` are venue-specific and unrelated.
- The `MyriadRawMarket` type is consumed by `normalizer.ts` and `utils.ts` -- both are updated.
- `mapMarketToUnified` in `utils.ts` is called from `mapQuestionToEvent` (same file, line 90). No callers from `index.ts` or other modules.
- No SDK-layer changes needed -- this fix is at the raw data ingestion layer, before the pipeline reaches OpenAPI or SDK shims.

### Findings

**Test regression (blocking):** The existing test fixture in `core/test/normalizers/exchange-normalizers-2.test.ts:414` still sets `questionId: 7` on the `MyriadRawMarket` fixture. Because the interface now has `[key: string]: unknown`, TypeScript won't flag this -- but the normalizer now reads `raw.eventId`, which is `undefined` in the fixture. The test at line 452 ("sets eventId from questionId") will fail:

```
Expected: '7'
Received: undefined
```

Fix: change the fixture from `questionId: 7` to `eventId: 7`, and update the test description from "sets eventId from questionId" to "sets eventId from eventId" or similar.

### PMXT Pipeline Check
- **Venue normalizer**: Updated (normalizer.ts + utils.ts)
- **OpenAPI schema**: Not affected (eventId is already in the UnifiedMarket schema)
- **SDK shims**: Not affected (no change to the unified output shape)

### Semver Impact
`patch` -- bug fix only, no public API changes.

### Risk
LOW with the test fix. The `eventId` field in the live Myriad API is currently always null per #556, so this change makes the code read the correct (albeit empty) field. No behavioral regression for end users.

PR #603, opened 2026-05-24
perf: replace O(n²) concat with push in kalshi fetcher

### LINKED ISSUE #343: Unbounded: core/src/exchanges/kalshi/fetcher.ts — O(n²) array.concat() with MAX_PAGES=1000
## Location
`core/src/exchanges/kalshi/fetcher.ts:427` (also `:465`)

## Code
```typescript
const MAX_PAGES = 1000;
const BATCH_SIZE = 200;

// fetchActiveEvents (line 408)
do {
    const events = data.events || [];
    allEvents = allEvents.concat(events);  // ← creates new array every iteration
    // ...
} while (cursor && page < MAX_PAGES);

// fetchAllWithStatus (line 465) — identical pattern
allEvents = allEvents.concat(events);
```

## Growth Pattern
`Array.prototype.concat()` allocates a brand-new array containing all previously accumulated elements plus the new page. It does not mutate in place. At page `k`, the new array contains `k × BATCH_SIZE` element references, all newly allocated slots. The previous array becomes garbage, but until GC collects it both arrays coexist in memory.

Total pointer slots allocated across all iterations: `BATCH_SIZE × (1 + 2 + 3 + … + N) = BATCH_SIZE × N(N+1)/2`

With `BATCH_SIZE=200` and `MAX_PAGES=1000`:
- Total pointer copies: `200 × (1000 × 1001 / 2)` = **100,100,000** pointer-sized allocations
- At 8 bytes/pointer: **~801 MB** in intermediate array allocations

Even with generational GC, the largest live array grows to 200,000 pointers (1.6 MB) while the second-largest is still alive, creating a sustained dual-allocation peak.

## OOM Estimate
- Worst case (1000 pages, all full): ~800 MB of intermediate arrays in flight during a single `fetchMarkets()` call
- Realistic Kalshi (50 pages × 200 events): ~200*(50*51/2) = 255,000 pointers ≈ **2 MB** peak overhead — manageable, but GC pressure spikes on every call
- If `fetchAllWithStatus('settled')` runs concurrently with `fetchActiveEvents()`, both do the same pattern simultaneously → peak memory doubles

## Suggested Fix
Replace `concat` with `push` spread: `allEvents.push(...events)`. Pre-allocating with `const allEvents: KalshiRawEvent[] = []` and using in-place push eliminates all intermediate array allocations. For very large pages, use `Array.prototype.push.apply(allEvents, events)` to avoid spreading a large array onto the call stack.

---

realfishsam:
Fixes #343

### DIFF
--- a/core/src/exchanges/kalshi/fetcher.ts
+++ b/core/src/exchanges/kalshi/fetcher.ts
@@ -474,7 +474,7 @@ export class KalshiFetcher implements IExchangeFetcher<KalshiRawEvent, KalshiRaw
                 const events = data.events || [];
                 if (events.length === 0) break;
 
-                allEvents = allEvents.concat(events);
+                allEvents.push(...events);
 
                 if (targetMarketCount) {
                     for (const event of events) {
@@ -512,7 +512,7 @@ export class KalshiFetcher implements IExchangeFetcher<KalshiRawEvent, KalshiRaw
             const events = data.events || [];
             if (events.length === 0) break;
 
-            allEvents = allEvents.concat(events);
+            allEvents.push(...events);
             cursor = data.cursor;
             page++;
         } while (cursor && page < MAX_PAGES);
@@ -546,7 +546,7 @@ export class KalshiFetcher implements IExchangeFetcher<KalshiRawEvent, KalshiRaw
             page++;
 
             if (events.length === 0) break;
-            allEvents = allEvents.concat(events);
+            allEvents.push(...events);
         } while (cursor && allEvents.length < maxEvents && page < MAX_PAGES);
 
         return {


realfishsam:
## PR Review: VERIFIED

### What This Does
Replaces three `allEvents = allEvents.concat(events)` calls with `allEvents.push(...events)` in the Kalshi fetcher's paginated event-fetching loops. Both produce the same result, but `push` mutates the existing array in-place instead of allocating a new copy on every page.

### Blast Radius
Minimal. Three lines changed in `core/src/exchanges/kalshi/fetcher.ts`. The call sites (`fetchEventsForMarkets`, `fetchAllWithStatus`, `fetchPageWithStatus`) are internal pagination loops where `allEvents` is a local `let` array -- the old code was already mutating the binding via reassignment, so switching to `push` is semantically equivalent and avoids O(n) copies per page.

### Findings
- **No issues found.** The change is correct and safe. `push(...spread)` is the standard pattern for accumulating into a local array.
- One edge-case to be aware of: if `events` ever contained tens of thousands of elements in a single page, `push(...events)` could hit the call-stack argument limit (~65k). Given `BATCH_SIZE` governs page size, this is not a realistic concern here.

### Semver Impact
**patch** -- internal perf improvement, no API surface change.

### Risk
Low.

PR #604, opened 2026-05-24
fix: opinion fetcher perf + websocket safety

### LINKED ISSUE #239: Missing timeout: Opinion WebSocket — no connection timeout on new WebSocket()
## Location
`core/src/exchanges/opinion/websocket.ts:67`

## Code
```typescript
this.connectionPromise = new Promise<void>((resolve, reject) => {
    try {
        this.ws = new WebSocket(this.wsUrl);

        this.ws.on("open", () => {
            this.isConnected = true;
            this.isConnecting = false;
            this.connectionPromise = undefined;
            resolve();
        });
```

## Risk
No connection timeout. If the Opinion WebSocket endpoint hangs at the TCP or TLS handshake stage, the `connectionPromise` never resolves or rejects, and all order book subscriptions block indefinitely.

## Affected Methods
- `OpinionWebSocket.connect()`
- `resubscribeAll()` and all subscription methods that depend on an open connection

## Suggested Fix
```typescript
const timeout = setTimeout(() => {
    ws.terminate();
    reject(new Error('Opinion WebSocket connection timeout'));
}, 15_000);

this.ws.on('open', () => { clearTimeout(timeout); resolve(); });
this.ws.on('error', (err) => { clearTimeout(timeout); reject(err); });
```

---
*Found by automated missing timeout audit*

### LINKED ISSUE #249: [non-null] opinion/websocket.ts: 2 unsafe Map.get()! assertions on resolver maps
## Risk Level
HIGH

## File
`core/src/exchanges/opinion/websocket.ts`

## Findings
- Line 323: `this.orderBookResolvers.get(marketId)!.push({ resolve, reject });` — assumes `marketId` is present in `orderBookResolvers`.
- Line 358: `this.tradeResolvers.get(marketId)!.push({ resolve, reject });` — assumes `marketId` is present in `tradeResolvers`.

`TypeError: Cannot read properties of undefined (reading 'push')` — the watcher promise for `watchOrderBook` or `watchTrades` leaks permanently. The subscriber is never notified of the failure; downstream code hangs until the watch timeout fires.

## Suggested Fix
Apply `has` + `set` guards before each push:
```typescript
if (!this.orderBookResolvers.has(marketId)) this.orderBookResolvers.set(marketId, []);
this.orderBookResolvers.get(marketId)!.push({ resolve, reject });
```

---

### LINKED ISSUE #347: Unbounded: core/src/exchanges/opinion/fetcher.ts — O(n²) spread operator with MAX_PAGES=500
## Location
`core/src/exchanges/opinion/fetcher.ts:541`

## Code
```typescript
const MAX_PAGES = 500;
const OPINION_MAX_PAGE_SIZE = 20;  // from config.ts

while (page <= MAX_PAGES) {
    // ...
    const list = extractList(result);
    allItems = [...allItems, ...list];  // ← new full-copy array every iteration
    // ...
    page += 1;
}
```

## Growth Pattern
The spread operator `[...allItems, ...list]` allocates a new array of length `allItems.length + list.length` on every iteration — it does not mutate in place. At iteration `k`, the new array has `k × OPINION_MAX_PAGE_SIZE` slots, and the old array (same size minus one page) is eligible for GC but may not yet be collected.

Total pointer slots allocated across all iterations:
`OPINION_MAX_PAGE_SIZE × (1 + 2 + … + MAX_PAGES) = 20 × (500 × 501 / 2)` = **2,505,000** pointer copies

With 500 pages at 20 items/page = 10,000 final items. The final array is 10,000 pointers but the intermediate allocations sum to 2.5 million pointer slots.

This `fetchPaginated` helper is used for all market/event listing in the Opinion exchange — `fetchMarketPages`, `fetchEventPages`, etc. — making it a hot path.

## OOM Estimate
- Each pointer: 8 bytes → ~20 MB of intermediate allocations per full `fetchMarkets()` call
- Each OpinionMarket item: ~500 bytes; 10,000 items = 5 MB for the actual data
- Intermediate array churn: 4× the final data size, causing repeated GC pauses
- Under concurrent requests (e.g., server handling 10 simultaneous `fetchMarkets` calls): 10 × 20 MB = **200 MB** of short-lived garbage in flight

## Suggested Fix
Replace `allItems = [...allItems, ...list]` with `allItems.push(...list)`. Since `list` is bounded by `OPINION_MAX_PAGE_SIZE = 20`, the spread-onto-call-stack is safe. This eliminates all intermediate array allocations.

---

realfishsam:
Fixes #239
Fixes #249
Fixes #347

### DIFF
--- a/core/src/exchanges/opinion/fetcher.ts
+++ b/core/src/exchanges/opinion/fetcher.ts
@@ -538,7 +538,7 @@ export class OpinionFetcher implements IExchangeFetcher<OpinionRawMarket, Opinio
 
             if (!list || list.length === 0) break;
 
-            allItems = [...allItems, ...list];
+            allItems.push(...list);
 
             if (allItems.length >= total) break;
             if (targetCount !== undefined && allItems.length >= targetCount) break;
--- a/core/src/exchanges/opinion/websocket.ts
+++ b/core/src/exchanges/opinion/websocket.ts
@@ -65,9 +65,29 @@ export class OpinionWebSocket {
 
     this.connectionPromise = new Promise<void>((resolve, reject) => {
       try {
+        const CONNECTION_TIMEOUT_MS = 30_000;
+        let settled = false;
+
+        const connectionTimer = setTimeout(() => {
+          if (!settled) {
+            settled = true;
+            this.isConnecting = false;
+            this.connectionPromise = undefined;
+            logger.error("Opinion WebSocket connection timed out", { timeoutMs: CONNECTION_TIMEOUT_MS });
+            if (this.ws) {
+              this.ws.terminate();
+              this.ws = undefined;
+            }
+            reject(new Error(`Opinion WebSocket connection timed out after ${CONNECTION_TIMEOUT_MS}ms`));
+          }
+        }, CONNECTION_TIMEOUT_MS);
+
         this.ws = new WebSocket(this.wsUrl);
 
         this.ws.on("open", () => {
+          if (settled) return;
+          settled = true;
+          clearTimeout(connectionTimer);
           this.isConnected = true;
           this.isConnecting = false;
           this.connectionPromise = undefined;
@@ -87,6 +107,9 @@ export class OpinionWebSocket {
         });
 
         this.ws.on("error", (error: Error) => {
+          if (settled) return;
+          settled = true;
+          clearTimeout(connectionTimer);
           logger.error("Opinion WebSocket error", { error: String(error) });
           this.isConnecting = false;
           this.connectionPromise = undefined;
@@ -318,10 +341,12 @@ export class OpinionWebSocket {
     }
 
     const dataPromise = new Promise<OrderBook>((resolve, reject) => {
-      if (!this.orderBookResolvers.has(marketId)) {
-        this.orderBookResolvers.set(marketId, []);
+      const existing = this.orderBookResolvers.get(marketId);
+      if (existing) {
+        existing.push({ resolve, reject });
+      } else {
+        this.orderBookResolvers.set(marketId, [{ resolve, reject }]);
       }
-      this.orderBookResolvers.get(marketId)!.push({ resolve, reject });
     });
 
     return withWatchTimeout(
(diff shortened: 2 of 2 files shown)


realfishsam:
## PR Review: VERIFIED

### What This Does
Two changes in Opinion exchange code:

1. **fetcher.ts**: Replaces `allItems = [...allItems, ...list]` (O(n^2) spread-copy per page) with `allItems.push(...list)` (amortized O(1) append). Same pattern as PR #603.

2. **websocket.ts**: Three improvements:
   - Adds a 30s connection timeout with `settled` flag to prevent the connection promise from hanging indefinitely if the WS server never responds.
   - Guards `open` and `error` handlers against double-settlement via the `settled` flag, and clears the timeout on settlement.
   - Removes two `Map.get()!` non-null assertions in `watchOrderBook` and `watchTrades` by capturing the result of `get()` into a local and branching.

### Blast Radius
Two files in `core/src/exchanges/opinion/`. The fetcher change is isolated to a single pagination loop. The websocket changes are confined to the `connect()` promise lifecycle and the two `watch*` resolver-push sites.

### Findings
- **Connection timeout is correctly implemented.** The `settled` flag prevents both the timer callback and the `open`/`error` handlers from double-settling the promise. `ws.terminate()` is correctly used (hard close, no graceful FIN) since the connection never completed.
- **Non-null assertion removal is clean.** The new pattern (`const existing = map.get(key); if (existing) { existing.push(...) } else { map.set(key, [...]) }`) avoids the intermediate empty-array allocation from the old `if (!has) set([])` + `get()!.push()` pattern.
- **Minor note:** The `close` event handler in `connect()` does not clear the timeout or check `settled`. If the WS closes before `open` fires (e.g., server rejects immediately via close frame without error), the timeout would still fire 30s later and call `reject()` on an already-resolved promise -- but `settled` protects against that, so this is safe. The `close` handler itself does not resolve or reject, it just schedules reconnect, so no double-settlement risk.

### Semver Impact
**patch** -- bug fix (timeout + assertion safety), no API change.

### Risk
Low.

PR #605, opened 2026-05-24
fix: limitless non-null assertions + resolver leak cleanup

### LINKED ISSUE #257: [non-null] limitless/websocket.ts: 4 unsafe assertions (Map.get()!, buffer.shift()!, resolver.shift()!)
## Risk Level
HIGH

## File
`core/src/exchanges/limitless/websocket.ts`

## Findings
- Line 119: `return buffer.shift()!;` — `Array.prototype.shift()` returns `T | undefined` when the array may be empty. If the buffer is empty at the call site, this returns `undefined` cast to `T`.
- Line 317: `const resolver = resolvers.shift()!;` — same as above on the resolvers array; if empty, `resolver` is `undefined`, and the next dereference of `resolver.resolve(...)` crashes.
- Line 324: `const buffer = this.orderbookBuffers.get(marketSlug)!;` — `Map.get()` returns `T | undefined`; if `marketSlug` is not in the map, all subsequent operations on `buffer` crash.
- Line 158: `this.orderbookResolvers.get(marketSlug)!.push({ resolve, reject });` — same `Map.get()!` pattern on the resolvers map.

- Line 119: Silent `undefined` return, likely causing a downstream `TypeError` when the caller tries to use the value.
- Line 317: `TypeError: Cannot read properties of undefined (reading 'resolve')` — resolver is lost, promise leaks.
- Lines 324/158: `TypeError: Cannot read properties of undefined (reading 'push'/'length'/etc.)` — orderbook state machine breaks for that `marketSlug`.

## Suggested Fix
- Lines 119/317: Guard with a length check before shifting, or assert the array is non-empty explicitly with a descriptive error.
- Lines 324/158: Use `has` + `set` guard before `get`:
```typescript
if (!this.orderbookBuffers.has(marketSlug)) {
    throw new Error(`[limitless] No orderbook buffer for ${marketSlug}`);
}
const buffer = this.orderbookBuffers.get(marketSlug)!;
```

---

### LINKED ISSUE #290: [non-null] limitless/client.ts: 3 unsafe assertions on orderClient and signer
## Risk Level
MEDIUM

## File
`core/src/exchanges/limitless/client.ts`

## Findings
- Line 311: `return await this.orderClient!.cancel(orderId);`
- Line 318: `return await this.orderClient!.cancelAll(marketSlug);`
- Line 388: `const balance = await contract.balanceOf(this.signer!.address);`

`orderClient` is initialized lazily (only when trading credentials are present); `signer` is set during wallet setup. Both are optional at construction time. Calling `cancel`, `cancelAll`, or `getBalance` on a read-only client instance (no private key provided) triggers the assertions.

`TypeError: Cannot read properties of undefined (reading 'cancel'/'cancelAll'/'address')` — the operation fails with a generic error instead of a helpful "trading credentials not configured" message.

## Suggested Fix
Guard each use site explicitly:
```typescript
if (!this.orderClient) throw new Error('[limitless] Order client not initialized — trading credentials required');
return await this.orderClient.cancel(orderId);
```
```typescript
if (!this.signer) throw new Error('[limitless] Signer not initialized — wallet private key required');
const balance = await contract.balanceOf(this.signer.address);
```

---

### LINKED ISSUE #303: [non-null] limitless/normalizer.ts: 2 params.start! and params.end! assertions inside guarded callbacks
## Risk Level
MEDIUM

## File
`core/src/exchanges/limitless/normalizer.ts`

## Findings
- Line 74: `candles = candles.filter((c) => c.timestamp >= params.start!.getTime());` — inside `if (params.start)`, but TypeScript does not narrow optional properties inside arrow-function callbacks.
- Line 77: `candles = candles.filter((c) => c.timestamp <= params.end!.getTime());` — same issue for `params.end`.

Logically safe: the `if (params.start)` / `if (params.end)` guards on lines 73/76 ensure the values are non-null before entering the block. TypeScript's control-flow analysis does not narrow through closures, so the `!` is needed to satisfy the compiler. No runtime crash is expected under current code.

The risk is a future refactor that removes the outer `if` guard without removing the `!` — the assertion silently hides the bug.

## Suggested Fix
Capture the value in a local `const` before the callback so TypeScript can narrow it:
```typescript
if (params.start) {
    const start = params.start;
    candles = candles.filter((c) => c.timestamp >= start.getTime());
}
if (params.end) {
    const end = params.end;
    candles = candles.filter((c) => c.timestamp <= end.getTime());
}
```

---

### LINKED ISSUE #372: Unbounded: core/src/exchanges/limitless/websocket.ts — stale orderbookResolvers accumulate when timeout races WS update
## Location
`core/src/exchanges/limitless/websocket.ts:154-176`

## Code
```typescript
// watchOrderBook — called in a streaming loop
const wsUpdatePromise = new Promise<OrderBook>((resolve, reject) => {
    if (!this.orderbookResolvers.has(marketSlug)) {
        this.orderbookResolvers.set(marketSlug, []);
    }
    this.orderbookResolvers.get(marketSlug)!.push({ resolve, reject }); // ← pushed here
});

const timeoutPromise = new Promise<OrderBook>((resolve) => {
    setTimeout(async () => {
        // Timeout fires — resolves the race winner
        const snapshot = await this.fetchOrderBookSnapshot(marketSlug);
        resolve(snapshot);
        // ↑ This resolves the Promise.race() but does NOT remove
        //   { resolve, reject } from orderbookResolvers
    }, WS_UPDATE_TIMEOUT); // 3 seconds
});

return await Promise.race([wsUpdatePromise, timeoutPromise]);
// After this returns, the resolver is STILL in orderbookResolvers.get(marketSlug)
```

In `setupEventHandlers` (line ~317):
```typescript
const resolvers = this.orderbookResolvers.get(marketSlug) || [];
if (resolvers.length > 0) {
    const resolver = resolvers.shift();  // ← only cleared on next WS update
    resolver.resolve(pmxtOrderbook);
} else {
    // buffered
}
```

## Growth Pattern
When `watchOrderBook()` races a 3-second timeout against a WS push, and the **timeout wins** (market is sparse or WS is slow), the `{ resolve, reject }` closure is left inside `orderbookResolvers.get(marketSlug)`. It is only evicted when the **next WS update** arrives and calls `resolvers.shift()`. If WS updates are infrequent (sparse prediction market), multiple timeout-winning calls can stack up:

- `watchOrderBook` called 10 times in 3 seconds, timeout fires each time → 10 stale closures in the array
- WS updates come once per minute → closures held for up to 60 seconds
- Each closure captures the `OrderBook` Promise resolver: ~1–2 KB of V8 closure state

Additionally, `close()` (line 277) calls `this.orderbookCallbacks.clear()` and `this.priceCallbacks.clear()` but does **not** clear `this.orderbookResolvers`. Stale resolvers from before `close()` are never freed.

## OOM Estimate
- At 100 markets × 10 concurrent watchers × 2 KB per closure = **2 MB** per polling cycle
- Under active streaming with 3s timeouts and 30s WS update intervals: up to 10 stale closures/market accumulate before each WS push clears them
- Sustained peak: 100 markets × 10 closures × 2 KB = **2 MB** perpetually held (not an OOM risk alone)
- Combined with many short-lived exchange instances: compounds with other leaks
- The `close()` gap means any shutdown leaves orphaned closures; if the caller retries with a new `LimitlessWebSocket` instance, old closures from the disposed instance are never freed

## Suggested Fix
1. In the timeout handler, remove the resolver from `orderbookResolvers` before resolving: use `findIndex`/`splice` to locate and remove the specific `{ resolve, reject }` object.
2. In `close()`, call `this.orderbookResolvers.clear()` after rejecting all pending resolvers.

---

realfishsam:
Fixes #257
Fixes #290
Fixes #303
Fixes #372

### DIFF
--- a/core/src/exchanges/limitless/client.ts
+++ b/core/src/exchanges/limitless/client.ts
@@ -309,14 +309,20 @@ export class LimitlessClient {
      * Cancel a specific order by ID.
      */
     async cancelOrder(orderId: string) {
-        return await this.orderClient!.cancel(orderId);
+        if (!this.orderClient) {
+            throw new Error('[limitless] Order client not initialized -- trading credentials required');
+        }
+        return await this.orderClient.cancel(orderId);
     }
 
     /**
      * Cancel all orders for a specific market.
      */
     async cancelAllOrders(marketSlug: string) {
-        return await this.orderClient!.cancelAll(marketSlug);
+        if (!this.orderClient) {
+            throw new Error('[limitless] Order client not initialized -- trading credentials required');
+        }
+        return await this.orderClient.cancelAll(marketSlug);
     }
 
     /**
@@ -385,7 +391,10 @@ export class LimitlessClient {
         });
         const contract = new Contract(USDC_ADDRESS, ABI, provider);
 
-        const balance = await contract.balanceOf(this.signer!.address);
+        if (!this.signer) {
+            throw new Error('[limitless] Signer not initialized -- wallet private key required');
+        }
+        const balance = await contract.balanceOf(this.signer.address);
         const decimals = await contract.decimals(); // Should be 6
 
         return parseFloat(utils.formatUnits(balance, decimals));
--- a/core/src/exchanges/limitless/normalizer.ts
+++ b/core/src/exchanges/limitless/normalizer.ts
@@ -71,10 +71,12 @@ export class LimitlessNormalizer implements IExchangeNormalizer<LimitlessRawMark
         }).sort((a, b) => a.timestamp - b.timestamp);
 
         if (params.start) {
-            candles = candles.filter((c) => c.timestamp >= params.start!.getTime());
+            const start = params.start;
+            candles = candles.filter((c) => c.timestamp >= start.getTime());
         }
         if (params.end) {
-            candles = candles.filter((c) => c.timestamp <= params.end!.getTime());
+            const end = params.end;
+            candles = candles.filter((c) => c.timestamp <= end.getTime());
         }
         if (params.limit) {
             candles = candles.slice(0, params.limit);
--- a/core/src/exchanges/limitless/websocket.ts
+++ b/core/src/exchanges/limitless/websocket.ts
@@ -117,7 +117,8 @@ export class LimitlessWebSocket {
         // 1. If we have buffered data, return it immediately
         const buffer = this.orderbookBuffers.get(marketSlug);
         if (buffer && buffer.length > 0) {
-            return buffer.shift()!;
+            const entry = buffer.shift();
+            if (entry) return entry;
         }
 
         // 2. Special case: If this is the FIRST call for this market and we have no data,
@@ -152,15 +153,31 @@ export class LimitlessWebSocket {
 
         // Wait for WebSocket update with timeout
         try {
+            const resolverEntry: { resolve: (ob: OrderBook) => void; reject: (err: any) => void } = {
+                resolve: () => {},
+                reject: () => {},
+            };
+
             const wsUpdatePromise = new Promise<OrderBook>((resolve, reject) => {
+                resolverEntry.resolve = resolve;
+                resolverEntry.reject = reject;
                 if (!this.orderbookResolvers.has(marketSlug)) {
                     this.orderbookResolvers.set(marketSlug, []);
                 }
-                this.orderbookResolvers.get(marketSlug)!.push({ resolve, reject });
+                const resolvers = this.orderbookResolvers.get(marketSlug);
+                if (resolvers) {
+                    resolvers.push(resolverEntry);
+                }
             });
 
             const timeoutPromise = new Promise<OrderBook>((resolve) => {
                 setTimeout(async () => {
+                    // Timeout won the race -- remove the stale resolver (#372)
+                    const resolvers = this.orderbookResolvers.get(marketSlug);
+                    if (resolvers) {
+                        const idx = resolvers.indexOf(resolverEntry);
+                        if (idx !== -1) resolvers.splice(idx, 1);
+                    }
                     // Timeout: fetch REST snapshot as fallback
                     try {
                         this.lastOrderbookTimestamps.set(marketSlug, Date.now());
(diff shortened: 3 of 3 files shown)


realfishsam:

### What This Does
Three files changed across the Limitless exchange:

1. **client.ts**: Replaces three `this.orderClient!` / `this.signer!` non-null assertions with explicit `if (!x) throw` guards in `cancelOrder`, `cancelAllOrders`, and `getBalance`. Error messages are clear and actionable.

2. **normalizer.ts**: Replaces two `params.start!` / `params.end!` non-null assertions inside their own `if (params.start)` / `if (params.end)` blocks with local const bindings. TypeScript narrows correctly when the value is captured before the closure.

3. **websocket.ts**: Multiple changes:
   - Removes `buffer.shift()!` by assigning to a local and checking.
   - Refactors the `watchOrderBook` resolver-push pattern to eliminate `Map.get()!` calls.
   - **Resolver leak fix (#372):** When the timeout wins the `Promise.race`, the stale resolver entry is now removed from the `orderbookResolvers` array via `indexOf` + `splice`. Previously, timed-out resolvers accumulated indefinitely.
   - **Close cleanup (#372):** `close()` now rejects all pending orderbookResolvers before clearing, and also clears `orderbookBuffers`.
   - Removes non-null assertions in `setupEventHandlers` for resolver shift and buffer push.

### Blast Radius
Three files in `core/src/exchanges/limitless/`. The client and normalizer changes are cosmetic safety improvements. The websocket resolver-leak fix and close-cleanup are behavioral changes that affect long-running connections.

### Findings
- **Resolver leak fix is correct.** The `resolverEntry` object is created outside the Promise constructor so it can be referenced by both the wsUpdatePromise and the timeout cleanup. `indexOf` identity comparison works because it is the same object reference.
- **`close()` now rejects pending resolvers** -- this is good. Without this, callers awaiting `watchOrderBook` would hang forever after `close()`.
- **Mutation concern in websocket.ts:** `resolvers.splice(idx, 1)` mutates the array in-place. This is acceptable here because the array is owned by the Map and only accessed from this module, but it is worth noting as a deviation from immutable patterns.
- **The `buffer.shift()` guard (line ~120):** After the `if (entry) return entry;` check, if `shift()` returned `undefined` (empty array that passed `length > 0` check -- impossible), execution falls through to the rest of the method. This is fine.
- **Not verified** because the websocket resolver-leak fix involves a race condition that requires runtime testing to confirm correctness under concurrent `watchOrderBook` calls. The logic reads correctly, but races are hard to verify statically.

### Semver Impact
**patch** -- bug fix (resolver leak, close cleanup), no API surface change.

### Risk
Medium. The resolver-leak fix changes race semantics in `watchOrderBook`. Incorrect cleanup could cause a resolver to be resolved after removal, or a timeout to fire on a cleaned-up resolver. The code looks correct, but integration testing is recommended.

PR #608, opened 2026-05-24
fix: guard Map.get() on candle buckets in polymarket normalizer

### LINKED ISSUE #321: [non-null] polymarket/normalizer.ts: unsafe Map.get()! on candle buckets
## Risk Level
MEDIUM

## File
`core/src/exchanges/polymarket/normalizer.ts`

## Findings
- Line 80: `const candle = buckets.get(snappedMs)!;` — placed in the `else` branch of `if (!buckets.has(snappedMs)) { buckets.set(...); } else { ... }`. Logically safe because the `else` guarantees `buckets.has(snappedMs)` is true. TypeScript does not track the `has`/`get` relationship, hence the `!`.

In practice this should never crash given the surrounding control flow. The risk is a future refactor that restructures the if/else without removing the `!`, allowing `candle` to be `undefined` while lines 81–84 attempt property assignments on it.

## Suggested Fix
Use a local variable narrowing to remove the assertion entirely:
```typescript
const candle = buckets.get(snappedMs);
if (candle) {
    candle.high = Math.max(candle.high, price);
    candle.low = Math.min(candle.low, price);
    candle.close = price;
    candle.volume = (candle.volume || 0) + volume;
}
```

---

realfishsam:
Fixes #321

### DIFF
--- a/core/src/exchanges/polymarket/normalizer.ts
+++ b/core/src/exchanges/polymarket/normalizer.ts
@@ -77,11 +77,13 @@ export class PolymarketNormalizer implements IExchangeNormalizer<PolymarketRawEv
                     volume: volume,
                 });
             } else {
-                const candle = buckets.get(snappedMs)!;
-                candle.high = Math.max(candle.high, price);
-                candle.low = Math.min(candle.low, price);
-                candle.close = price;
-                candle.volume = (candle.volume || 0) + volume;
+                const candle = buckets.get(snappedMs);
+                if (candle) {
+                    candle.high = Math.max(candle.high, price);
+                    candle.low = Math.min(candle.low, price);
+                    candle.close = price;
+                    candle.volume = (candle.volume || 0) + volume;
+                }
             }
         }
 


realfishsam:
## PR Review: VERIFIED

### What This Does
Guards `Map.get(snappedMs)` with an `if (candle)` check instead of using the non-null assertion `buckets.get(snappedMs)!` in the Polymarket normalizer's OHLCV candle-bucketing logic.

### Blast Radius
Single file: `core/src/exchanges/polymarket/normalizer.ts`. One code path in `normalizeOHLCV`. The `else` branch is only entered when `buckets.has(snappedMs)` is true (the preceding `if (!buckets.has(snappedMs))` sets the entry), so `get()` will always return a value in practice. The guard is purely defensive.

### Findings
- **The guard is correct and safe.** If `get()` somehow returned `undefined` (impossible given the `has()` check above, but defensive), the price point is silently skipped. This is acceptable for OHLCV data -- a missing candle update is better than a crash.
- **Mutation concern:** The candle object retrieved from the Map is still mutated in-place (`candle.high = ...`, `candle.low = ...`, etc.). This is a pre-existing pattern and not introduced by this PR. Since the Map owns these objects and they are only used locally within this method before being returned as an array, in-place mutation is acceptable here.
- **No non-null assertions remain** in this code path.

### Semver Impact
**patch** -- defensive guard, no behavioral change in practice.

### Risk
Low.

PR #609, opened 2026-05-24
fix: replace console.warn with structured logger in SDK router

### LINKED ISSUE #396: console: sdks/typescript/pmxt/router.ts:207 — console.warn for deprecated fetchMatches in SDK
## Violation
```typescript
console.warn('[pmxt] fetchMatches is deprecated, use fetchMarketMatches instead');
```

## Location
`sdks/typescript/pmxt/router.ts` line 207

## Why It Matters
SDK code should not use `console.warn` for deprecation notices — this bypasses any logger the consuming application has configured and pollutes their stderr unconditionally. SDK deprecation warnings should use a controlled mechanism.

## Suggested Fix
Consider using Node.js's `process.emitWarning()` or a configurable logger injected by the consumer. At minimum, replace with a logger so the emission can be suppressed.

---

realfishsam:
Fixes #396

### DIFF
--- a/sdks/typescript/pmxt/logger.ts
+++ b/sdks/typescript/pmxt/logger.ts
@@ -1,0 +1,78 @@
+/**
+ * Structured logger for pmxt TypeScript SDK.
+ *
+ * Thin abstraction over console that:
+ * - Attaches a `[pmxt]` prefix for easy filtering
+ * - Supports log levels (debug, info, warn, error)
+ * - Respects a configurable level threshold via PMXT_LOG_LEVEL
+ * - Can be swapped for a real transport later
+ */
+
+export type LogLevel = 'debug' | 'info' | 'warn' | 'error' | 'silent';
+
+const LEVEL_PRIORITY: Record<LogLevel, number> = {
+    debug: 0,
+    info: 1,
+    warn: 2,
+    error: 3,
+    silent: 4,
+};
+
+function getDefaultLevel(): LogLevel {
+    if (typeof process !== 'undefined' && process.env?.PMXT_LOG_LEVEL) {
+        return process.env.PMXT_LOG_LEVEL as LogLevel;
+    }
+    return 'info';
+}
+
+let currentLevel: LogLevel = getDefaultLevel();
+
+function shouldLog(level: LogLevel): boolean {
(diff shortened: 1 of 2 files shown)


realfishsam:

### What This Does
Adds a new structured logger module (`sdks/typescript/pmxt/logger.ts`) to the TypeScript SDK and replaces a single `console.warn` call in `router.ts` with `logger.warn`.

### Blast Radius
- **New file:** `sdks/typescript/pmxt/logger.ts` (78 lines). Introduces a `logger` singleton with `debug/info/warn/error` methods, a `[pmxt]` prefix, and level gating via `PMXT_LOG_LEVEL` env var.
- **Modified file:** `sdks/typescript/pmxt/router.ts` -- one line changed in the deprecated `fetchMatches` method.

### Findings
- **Module-level mutable state:** `let currentLevel: LogLevel = getDefaultLevel()` is module-scoped mutable state. `logger.setLevel()` mutates it. This means the log level is global across all `Exchange`/`Router` instances in the same process. This is standard for loggers but worth noting.
- **`getDefaultLevel()` reads `process.env` at module load time.** If `PMXT_LOG_LEVEL` is set after import, the initial level won't reflect it. Users can call `logger.setLevel()` to override. This is acceptable.
- **No validation on `PMXT_LOG_LEVEL`.** If the env var is set to an invalid value (e.g., `PMXT_LOG_LEVEL=verbose`), it will be cast to `LogLevel` without validation. `shouldLog()` will then look up `LEVEL_PRIORITY[currentLevel]` which returns `undefined`, and `undefined >= LEVEL_PRIORITY[level]` is `false` -- so all logging would be silently suppressed. Consider adding validation in `getDefaultLevel()`.
- **`typeof process !== 'undefined'` guard** enables browser compatibility. Good.
- **Only one call site updated.** The `console.warn` in `router.ts` is the only usage. If the goal is to replace all `console.*` calls in the SDK, this is incomplete but could be intentional as a first step.
- **Not verified** because the new logger module introduces a new public export (`logger`) that SDK consumers could import. Need to confirm this aligns with the SDK's public API surface and doesn't interfere with the generated client code.

### Semver Impact
**minor** -- new `logger` export added to the SDK package, no breaking changes.

### Risk
Low-Medium. The missing env var validation is the main concern -- an invalid `PMXT_LOG_LEVEL` silently kills all logging.

PR #683, opened 2026-05-26
fix: avoid float precision loss in Limitless balances

### LINKED ISSUE #675: Float safety: Limitless balance parseFloat(formatUnits) and parseFloat / Math.pow float conversions
## Risk Level
HIGH

## Location
`core/src/exchanges/limitless/client.ts:400`
`core/src/exchanges/limitless/index.ts:585`

## Code
```typescript
// client.ts:400 — on-chain USDC balance fetch
return parseFloat(utils.formatUnits(balance, decimals));

// index.ts:585 — alternate balance path
const total = parseFloat(rawBalance.toString()) / Math.pow(10, USDC_DECIMALS);
```

## Problem
Both paths convert a raw on-chain balance (bigint lamport/wei value) to a human-readable float amount, then use `parseFloat` as the last step.

**client.ts:400**: `utils.formatUnits(balance, decimals)` (from ethers.js/viem) produces a precise decimal string (e.g. `"1234.567890"`). `parseFloat()` on that string introduces IEEE 754 rounding for any value with more significant digits than IEEE 754 can represent exactly. The result is returned as the user's USDC balance.

**index.ts:585**: `rawBalance.toString()` → `parseFloat()` → `/ Math.pow(10, USDC_DECIMALS)` is a two-step error: `parseFloat` on the raw integer string can lose precision for values > 2^53, then `/ 1e6` float division adds a second rounding error. This is the same pattern as Myriad issue #256 (`Number(wei) / 1e18`).

USDC has 6 decimals, so `rawBalance` values up to `9.007e15 / 10^6 = 9.007e9` USDC stay within 2^53. For typical users this is safe, but programmatic market makers with large positions could exceed the threshold.

The balance is used to compute available funds for order sizing.

## Example Failure
```javascript
// client.ts:400 — formatUnits returns long decimal string
utils.formatUnits(1234567890123n, 6) = "1234567.890123"
parseFloat("1234567.890123") = 1234567.890123  // exact for this case

// But for very large balances:
utils.formatUnits(9999999999999999n, 6) = "9999999999.999999"
parseFloat("9999999999.999999") = 10000000000  // wrong! > 2^53/1e6 precision limit

// index.ts:585 — raw bigint too large for Number:
rawBalance = 9007199254740993n  // 2^53 + 1
parseFloat("9007199254740993") = 9007199254740992  // loses last digit
/ 1e6 = 9007199254.740992  // balance is off by 1 micro-USDC
```

## Suggested Fix
```typescript
// client.ts:400 — keep as string until the last moment:
return new Decimal(utils.formatUnits(balance, decimals)).toNumber();

// index.ts:585 — use BigInt division:
const total = Number((rawBalance * 10000n) / BigInt(Math.pow(10, USDC_DECIMALS))) / 10000;
// or with decimal.js:
import Decimal from 'decimal.js';
const total = new Decimal(rawBalance.toString()).dividedBy(new Decimal(10).pow(USDC_DECIMALS)).toNumber();
```

---
*Found by automated float safety audit*

00anon0X:
Fixes #675.

This avoids precision loss when converting raw USDC integer balances for Limitless.

Changes:
- Adds a shared `scaledIntegerToNumber` helper using integer division/modulo before final number conversion.
- Replaces unsafe balance conversions using `parseFloat(utils.formatUnits(...))` and `parseFloat(rawBalance.toString()) / Math.pow(...)`.
- Adds unit coverage for a raw balance above `Number.MAX_SAFE_INTEGER` and ethers `BigNumber` input.

Verification:
- `npm --workspace=pmxt-core test -- --runInBand core/test/unit/limitless-balance.core.test.ts`
- `npm --workspace=pmxt-core run build`


realfishsam (commented):
## PR Review: VERIFIED

### What This Does
Adds a `scaledIntegerToNumber(value, decimals)` helper in `limitless/utils.ts` that performs integer division/modulo in bigint arithmetic before converting to float, eliminating the precision loss that occurs when `rawBalance > Number.MAX_SAFE_INTEGER` (≈ 9 quadrillion USDC raw units). Replaces two unsafe conversion call sites: `parseFloat(rawBalance.toString()) / Math.pow(10, USDC_DECIMALS)` in `index.ts` and `parseFloat(utils.formatUnits(balance, decimals))` in `client.ts`.

### Blast Radius
- **Limitless only** — two internal methods (`LimitlessExchange.fetchBalance` in `index.ts` and `LimitlessClient.getBalance` in `client.ts`)
- No changes to unified types, OpenAPI schema, SDK clients, or other exchanges
- The helper is not exported from the exchange `index.ts`, so it doesn't widen the public API surface

---

### Consumer Verification

**Before (base branch — `parseFloat(rawBalance.toString()) / Math.pow(10, USDC_DECIMALS)`):**
```
raw = 9007199254740993n  (> Number.MAX_SAFE_INTEGER)
parseFloat("9007199254740993")  →  9007199254740992  (rounds, loses the last digit)
9007199254740992 / 1_000_000    →  9007199254.740992  ← WRONG (off by 1 ulp in the 6th decimal)
```
`fetchBalance` would return `9007199254.740992 USDC` instead of `9007199254.740993 USDC`.

**After (PR branch — `scaledIntegerToNumber`):**
```
raw = 9007199254740993n
whole   = 9007199254740993n / 1000000n  =  9007199254n  (exact bigint division)
fraction= 9007199254740993n % 1000000n  =    740993n    (exact bigint remainder)
result  = Number(9007199254) + Number(740993)/Number(1000000)
        = 9007199254 + 0.740993  =  9007199254.740993  ← CORRECT
```

Both the `index.ts` path (native bigint from on-chain call) and the `client.ts` path (ethers v5 BigNumber via `toBigInt()` / `toString()` fallback) are fixed. Confirmed in Node.js:
```
BEFORE (parseFloat/pow):        9007199254.740992   WRONG (off by 1 ulp)
AFTER  (scaledIntegerToNumber): 9007199254.740993   CORRECT
BigNumber toString fallback:    1234567.890123      CORRECT
decimals=0 edge case:           42                  CORRECT
```

Server starts cleanly on PR branch; `POST /api/limitless/fetchBalance` is reachable (returns expected auth error without credentials, as intended).

---

### Test Results
- **Build:** PASS (tsc clean)
- **New unit tests (`limitless-balance.core.test.ts`):** PASS (2/2)
- **Full suite on PR branch:** 3 suites failing, 53 tests failing
- **Full suite on main:** 4 suites failing, 57 tests failing

The 3 failing suites on the PR branch (`exchange-normalizers`, `param-forwarding`, `watch-order-book-api`) are **identical to failures on main** — pre-existing, unrelated to this change. The PR branch actually resolves a 4th suite (`schema-drift.test.ts`) that fails on current main, because the PR is based on an older commit (`389e781`) that predates the drift regression on main.

- **E2E smoke (Limitless fetchMarkets):** PASS — live markets returned with correct shape

---

### Findings

1. **`Number(decimals)` in `client.ts:401` relies on implicit toString() coercion** — `contract.decimals()` returns an ethers v5 BigNumber; `Number(bigNumber)` falls through to `toString()` (returns "6") then converts, giving 6. This works and is validated by `scaledIntegerToNumber`'s guard, but `Number(decimals.toString())` or `decimals.toNumber()` would be more explicit. Not blocking.

2. **`index.ts:316–320` still contains `parseFloat(m.matchedSize) ... / Math.pow(10, USDC_DECIMALS)` for order fill sizes** — out of scope for this PR (those values originate as API string floats, not raw on-chain integers, so the overflow path via `> MAX_SAFE_INTEGER` is different). Not a regression introduced here.

3. **`mergeable_state: "dirty"`** — branch needs a rebase onto current main before it can be merged. Unrelated to correctness of the fix.

---

### PMXT Pipeline Check
- Field propagation (3-layer): N/A — no new fields
- OpenAPI sync: N/A — no schema changes
- Financial precision: **OK** — this PR is precisely fixing a financial precision bug
- Type safety: OK — no `any` introduced, no `!` assertions, generic type union handles both `bigint` and ethers BigNumber
- Auth safety: N/A

---

### Semver Impact
**patch** — pure bug fix to an internal balance conversion helper, no API surface change

### Risk
On-chain balance calls require real credentials on Base mainnet, so the exact balance return value at the `fetchBalance` API boundary was verified via unit test and math proof rather than a live end-to-end call. The negative-balance branch in `scaledIntegerToNumber` is unreachable in practice (ERC-20 `balanceOf` returns uint256), but the guard is harmless.

---

PR #774, opened 2026-05-31
fix(python-sdk): rename Suibets to SuiBets for cross-SDK consistency

### LINKED ISSUE #773: [core-sdk-gap] Python SDK exports exchange class as `Suibets` instead of `SuiBets`
## Gap

The SuiBets exchange class is named `SuiBets` everywhere except the Python SDK, which uses `Suibets` (lowercase `b`). This breaks naming consistency across the platform and makes it impossible to write exchange-agnostic code that references the class by its canonical name.

## Core

The exchange implementation is registered under the identifier `suibets` and exposes the class `SuiBetsExchange`. The canonical public name across the platform is `SuiBets`.

## TypeScript SDK

**File:** `sdks/typescript/pmxt/client.ts`, line 2833

```typescript
export class SuiBets extends Exchange { ... }
```

Correct capitalisation.

## Python SDK

**File:** `sdks/python/pmxt/_exchanges.py`, line 473

```python
class Suibets(Exchange):
```

**File:** `sdks/python/pmxt/__init__.py`, lines 22 and 166

```python
from ._exchanges import Suibets
...
__all__ = [..., "Suibets", ...]
```

The class is defined and exported as `Suibets` — the `B` in `Bets` is lowercase.

## Expected behaviour

The Python SDK class and its export should be renamed to `SuiBets` to match the TypeScript SDK and core convention:

```python
# _exchanges.py
class SuiBets(Exchange): ...

# __init__.py
from ._exchanges import SuiBets
```

A deprecation alias (`Suibets = SuiBets`) may be warranted during a transition period to avoid breaking existing callers.

PrettyFox0:
Closes #773

## Summary
The TypeScript SDK and core engine both use `SuiBets` as the canonical class name. The Python SDK was the odd one out with the lowercase-b `Suibets`, breaking exchange-agnostic naming consistency across SDKs.

## Changes
- `sdks/python/pmxt/_exchanges.py`: rename class `Suibets` → `SuiBets`; keep `Suibets = SuiBets` deprecation alias so existing callers don't break.
- `sdks/python/pmxt/__init__.py`: import and export both names; mark `Suibets` in `__all__` as deprecated alias.

The `exchange_name="suibets"` registration string is unchanged (matches the issue's "registered identifier" note).

## Test plan
- [x] \`python -m py_compile sdks/python/pmxt/_exchanges.py sdks/python/pmxt/__init__.py\` passes
- [ ] \`from pmxt import SuiBets\` works for new callers
- [ ] \`from pmxt import Suibets\` still works for existing callers

### DIFF
--- a/core/scripts/generate-python-exchanges.js
+++ b/core/scripts/generate-python-exchanges.js
@@ -34,6 +34,12 @@ const OVERRIDES = {
             wallet_address: 'Wallet address (required for positions and balance)',
         },
     },
+    suibets: {
+        // Exchange name has no separator; force the canonical PascalCase used
+        // by the TypeScript SDK and core engine. The default-derived "Suibets"
+        // is emitted as a backwards-compatible alias.
+        className: 'SuiBets',
+    },
 };
 
 function toClassName(name) {
@@ -41,6 +47,14 @@ function toClassName(name) {
         .split(/[-_]/)
         .map(part => part.toLowerCase() === 'us' ? 'US' : part.charAt(0).toUpperCase() + part.slice(1))
         .join('');
+}
+
+// Resolve the canonical Python class name for an exchange wire key, honoring
+// any explicit override in OVERRIDES[name].className. Use this instead of
+// calling toClassName() directly so overrides apply everywhere consistently.
+function classNameFor(name) {
+    const ov = OVERRIDES[name] || {};
+    return ov.className || toClassName(name);
 }
 
 function toLegacyClassName(name) {
@@ -111,7 +125,7 @@ function build(name, block) {
 
 function generateClass(exchange) {
     const { name, creds } = exchange;
-    const className = toClassName(name);
+    const className = classNameFor(name);
     const ov = OVERRIDES[name] || {};
     const aliases = ov.paramAliases || {};
     const defaults = ov.defaults || {};
@@ -226,7 +240,7 @@ function generateClass(exchange) {
 const appTs = fs.readFileSync(APP_TS_PATH, 'utf8');
 const exchanges = parseExchanges(appTs);
 const legacyAliases = exchanges
-    .map(ex => ({ legacyName: toLegacyClassName(ex.name), className: toClassName(ex.name) }))
+    .map(ex => ({ legacyName: toLegacyClassName(ex.name), className: classNameFor(ex.name) }))
     .filter(alias => alias.legacyName !== alias.className);
 
 const header = [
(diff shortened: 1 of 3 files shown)


realfishsam (commented):
## PR Review: FAIL

### What This Does
Renames the Python SDK's `Suibets` class to `SuiBets` to match the TypeScript SDK and improves cross-SDK naming consistency, keeping `Suibets` as a backwards-compatible alias. The motivation is valid — `client.ts` in the TypeScript SDK does define `export class SuiBets extends Exchange`.

### Blast Radius
Python SDK only. Two files changed: `sdks/python/pmxt/_exchanges.py` and `sdks/python/pmxt/__init__.py`. No core, no TypeScript SDK, no router, no OpenAPI schema.

### Consumer Verification
**Before (base branch):**
```python
from pmxt import SuiBets  # ImportError
from pmxt import Suibets   # OK
```

**After (PR branch):**
```python
from pmxt import SuiBets   # OK
from pmxt import Suibets   # OK (alias)
```

Cannot run end-to-end consumer verification — no CI check runs recorded on this PR (0 check runs returned). Build/test status is unknown.

### Test Results
- Build: UNKNOWN (no check runs)
- Unit tests: UNKNOWN (no check runs)
- Server starts: NOT TESTED
- E2E smoke: NOT TESTED

### Findings

**1. [BLOCKING] Hand-editing an auto-generated file.**

`sdks/python/pmxt/_exchanges.py` carries an explicit header at line 1:
```
# This file is auto-generated by core/scripts/generate-python-exchanges.js
# Do not edit manually.
# To regenerate: npm run generate:sdk:all --workspace=pmxt-core
```

The generator's `toClassName()` function in `core/scripts/generate-python-exchanges.js` produces `'Suibets'` for the exchange name `'suibets'`:
```js
function toClassName(name) {
    return name
        .split(/[-_]/)
        .map(part => part.toLowerCase() === 'us' ? 'US' : part.charAt(0).toUpperCase() + part.slice(1))
        .join('');
}
// toClassName('suibets') => 'Suibets'
```

The next run of `npm run generate:sdk:all` will overwrite `_exchanges.py` and the import/`__all__` sections of `__init__.py`, reverting both files to `Suibets`. The fix will silently disappear. This is not a hypothetical: the generator also manages the `__init__.py` import line and `# Exchanges` block in `__all__`, both of which this PR touches.

**The correct fix is to update the generator itself.** One way: add a `classNameOverride` field to the `OVERRIDES` map in `generate-python-exchanges.js` and thread it through `generateClass()`. Another: special-case the `suibets` exchange ID in `toClassName()`. Either approach must also emit the `Suibets = SuiBets` backwards-compat alias in the generated output (mirroring how `Polymarket_us = PolymarketUS` is emitted for `polymarket_us`).

**2. [Minor] `Suibets` alias emits no `DeprecationWarning`.**

`Suibets = SuiBets` in `_exchanges.py` and the `__all__` comment note it as deprecated, but accessing `pmxt.Suibets` at runtime is completely silent. If the intent is to eventually remove it, callers won't know to migrate. Consider a module-level `__getattr__` in `__init__.py` that issues `DeprecationWarning` when `Suibets` is accessed by name. This is a minor concern, but it matters if there's a timeline for removal.

### PMXT Pipeline Check
- Field propagation (3-layer): N/A — no data model changes
- OpenAPI sync: N/A — no BaseExchange/method changes
- Financial precision: N/A
- Type safety: N/A — pure rename/alias
- Auth safety: N/A

### Semver Impact
`patch` — adds a public alias, existing `Suibets` callers are unaffected. But: the fix doesn't survive regeneration, so it shouldn't ship until the generator is corrected.

### Risk
The fix is applied at the wrong layer. `_exchanges.py` will be overwritten by the generator on the next SDK release cycle, silently reintroducing the bug. The PR should be reworked to fix `core/scripts/generate-python-exchanges.js` (and optionally update `__init__.py` by re-running the generator) rather than patching the generated output directly.

---

PR #870, opened 2026-06-05
fix: surface websocket and feed hygiene issues

### LINKED ISSUE #275: Float safety: Gemini Titan WebSocket applyDelta tolerance comparison Math.abs(l.price - price) < 0.0001
## Risk Level
MEDIUM

## Location
`core/src/exchanges/gemini-titan/websocket.ts:211`

## Code
```typescript
private applyDelta(levels: OrderLevel[], price: number, size: number, sortOrder: 'asc' | 'desc'): void {
    const idx = levels.findIndex(l => Math.abs(l.price - price) < 0.0001);
    ...
}
```

## Problem
Same class of problem as the Kalshi tolerance comparison. A hardcoded `0.0001` threshold is used to find existing price levels instead of exact equality. This tolerance is equal to 1 basis point. For prediction market prices near 0.01 or 0.99 (common in highly-skewed markets), the relative error is enormous.

If the Gemini Titan exchange uses tick sizes smaller than 0.0001, two distinct levels will collide. If float rounding produces a delta with the same price integer as an existing level but outside the 0.0001 band (unlikely but possible after price inversion), the level is not found and a duplicate is inserted, corrupting the order book.

## Example Failure
```
Market tick size = 0.0001 (1 basis point)
Existing level at 0.1500
Delta arrives for 0.1501
Math.abs(0.1500 - 0.1501) = 0.0001 < 0.0001 → false (works)

But after float inversion (1 - noPrice):
rawNoPrice = 0.8499
1 - 0.8499 = 0.15009999999999997  (IEEE 754)
Math.abs(0.1500 - 0.15009999999999997) ≈ 9.99e-5 < 0.0001 → TRUE
→ delta applied to 0.1500 instead of 0.1501
```

## Suggested Fix
Store integer-keyed price levels in a `Map` and use exact integer equality. Avoid tolerance-based matching entirely.

---
*Found by automated float safety audit*

### LINKED ISSUE #678: Unhandled async: kalshi/websocket.ts — connect() swallow comment is incorrect; synchronous failure bypasses scheduleReconnect
## Risk Level
MEDIUM

## Location
`core/src/exchanges/kalshi/websocket.ts:483–487`, `522–524`, `568–570`

## Code
All three call sites follow the same pattern, e.g. in `watchOrderBook`:

```typescript
if (!this.isConnected) {
    this.connect().catch(() => {
        // Connection failed — scheduleReconnect is already queued via the
        // close handler, so we intentionally swallow here. The pending
        // resolver will be resolved when data arrives on the next connection.
    });
}
```

## What Happens on Failure
The comment's claim — *"scheduleReconnect is already queued via the close handler"* — is only true when the WebSocket is **successfully created** and then closed. If `connect()` fails **synchronously** (before the WebSocket object is created), the `close` event never fires:

```typescript
// inside connect():
this.connectionPromise = new Promise((resolve, reject) => {
    try {
        const headers = this.auth.getHeaders("GET", path); // ← can throw (bad key)
        this.ws = new WebSocket(this.wsUrl, { headers });
        this.ws.on("close", () => {
            // scheduleReconnect() is only reachable from here
            this.scheduleReconnect();
        });
    } catch (error) {
        reject(error); // ← scheduleReconnect is NEVER called
    }
});
```

Concrete triggers:
- `KalshiAuth.getHeaders()` throws on a malformed or expired private key.
- Any synchronous exception in the constructor body.

When this happens:
1. `connect()` rejects.
2. The `.catch(() => {})` swallows the rejection.
3. `scheduleReconnect()` is never scheduled — the reconnect loop is dead.
4. The pending orderbook/trade resolvers sit in their Maps until `withWatchTimeout` fires (default 30 s).
5. **If `watchTimeoutMs: 0` (documented "no timeout"), the resolvers hang indefinitely.**

## Suggested Fix

At minimum, log the error in the swallow so it is visible:

```typescript
this.connect().catch((err: unknown) => {
    // Synchronous constructor failure — scheduleReconnect may not have been
    // queued. Log and schedule manually so the reconnect loop stays alive.
    logger.warn('KalshiWebSocket: connect() failed', {
    });
    this.scheduleReconnect();
});
```

The same fix applies to the `watchOrderBooks` (line 522) and `watchTrades` (line 568) swallows.

---

### LINKED ISSUE #679: Unhandled async: opinion/websocket.ts — connect() swallow has no logging and bypasses scheduleReconnect on synchronous failure
## Risk Level
MEDIUM

## Location
`core/src/exchanges/opinion/websocket.ts:336–338`, `373–375`

## Code
`watchOrderBook` (line 336) and `watchTrades` (line 373) both use:

```typescript
if (!this.isConnected) {
    this.connect().catch(() => {
        // Swallow — scheduleReconnect will retry. Resolvers stay pending.
    });
}
```

## What Happens on Failure
This is structurally identical to the Kalshi issue (#678). If `connect()` fails **before the WebSocket object is created** (a synchronous throw inside the `connect()` `try` block), the `'close'` event never fires and `scheduleReconnect()` is never called — contradicting the comment "scheduleReconnect will retry."

```typescript
// inside connect():
this.connectionPromise = new Promise<void>((resolve, reject) => {
    try {
        this.ws = new WebSocket(this.wsUrl);
        this.ws.on("close", () => {
            this.scheduleReconnect(); // ← only reachable if ws was created
        });
    } catch (error) {
        reject(error); // ← scheduleReconnect never runs
    }
});
```

Concrete triggers:
- `new WebSocket(this.wsUrl)` throws (e.g. invalid URL format, missing `ws://` scheme).
- Any synchronous exception in the constructor body.

When this happens:
1. `connect()` rejects.
2. `.catch(() => {})` swallows the rejection silently — no log line.
3. The reconnect loop never starts.
4. Pending orderbook/trade resolvers hang until `withWatchTimeout` (default 30 s).
5. **With `watchTimeoutMs: 0`, resolvers hang indefinitely.**

Unlike the Kalshi case, the swallow here has *no comment explaining the intent* — it is indistinguishable from an accidental empty catch.

## Suggested Fix

Log the error and explicitly schedule reconnect to keep the loop alive:

```typescript
this.connect().catch((err: unknown) => {
    logger.warn('OpinionWebSocket: connect() failed', {
    });
    this.scheduleReconnect();
});
```

Apply to both the `watchOrderBook` (line 336) and `watchTrades` (line 373) call sites.

---

### LINKED ISSUE #686: Unbounded: core/src/feeds/binance/binance-feed.ts — new Map() reconstructed on every WebSocket message in handleMessage
## Location
`core/src/feeds/binance/binance-feed.ts:225`

## Code
```typescript
private handleMessage(data: WebSocket.Data): void {
    // ...
    const ticker = normalizeTradeToTicker(event);

    // ← Creates a brand-new Map on EVERY incoming trade message
    this.latestTickers = new Map(this.latestTickers).set(ticker.symbol, ticker);

    for (const sub of this.subscriptions) {
        if (sub.symbol === ticker.symbol) {
            sub.callback(ticker);
        }
    }
}
```

## Growth Pattern
The Binance relay is a high-frequency trade firehose. Every trade event (potentially thousands per second across BTC/ETH/SOL/XRP) triggers `handleMessage()`, which:

1. Constructs a brand-new `Map` by copying all existing entries from `this.latestTickers`.
2. Calls `.set()` to add/update the single new ticker.
3. Discards the old Map (making it eligible for GC).

This is not an OOM issue per se (the Map size stays bounded at ~4 symbols), but it creates **continuous GC churn** at message-rate frequency:

- Binance spot: ~1000–5000 trades/sec across 4 symbols
- Each message allocates + immediately discards a ~4-entry Map object (~400 bytes overhead)
- GC must collect 1000–5000 short-lived Map objects per second
- This causes frequent minor GC pauses degrading streaming latency

The same pattern exists identically in `ChainlinkFeed.handleMessage()` at `core/src/feeds/chainlink/chainlink-feed.ts:396`.

## OOM Estimate
Not a direct OOM risk, but at 5000 msg/s:
- ~2 MB/s of short-lived Map allocations → V8 GC triggered every few hundred ms
- Under sustained load, GC pause frequency increases, causing jitter in the streaming callbacks that deliver tickers to WebSocket clients
- In the worst case (GC throughput bottleneck), the allocator cannot reclaim fast enough and heap grows until OOM

## Suggested Fix
Mutate in place instead of rebuilding the Map on every message:
```typescript
// Before:
this.latestTickers = new Map(this.latestTickers).set(ticker.symbol, ticker);

// After:
this.latestTickers.set(ticker.symbol, ticker);
```

The Map is only read (never iterated concurrently from another thread in Node.js single-threaded context), so in-place mutation is safe.

Apply the same fix to `ChainlinkFeed.handleMessage()` at `core/src/feeds/chainlink/chainlink-feed.ts:396`.

---

### LINKED ISSUE #687: Unbounded: core/src/feeds/binance/binance-feed.ts + chainlink-feed.ts — O(n²) array spread on every watchTickerImpl call
## Location
`core/src/feeds/binance/binance-feed.ts:133-134` (`watchTickerImpl`)
`core/src/feeds/chainlink/chainlink-feed.ts:181-182` (`watchTickerImpl`)

## Code
```typescript
// BinanceFeed.watchTickerImpl (line 133)
protected watchTickerImpl(symbol: string, callback: (ticker: Ticker) => void): () => void {
    const sub: Subscription = { symbol, callback };
    this.subscriptions = [...this.subscriptions, sub];  // ← O(n) copy on every subscribe

    return () => {
        this.subscriptions = this.subscriptions.filter((s) => s !== sub);  // ← O(n) copy on every unsubscribe
    };
}

// ChainlinkFeed.watchTickerImpl (line 181) — identical pattern
protected watchTickerImpl(symbol: string, callback: (ticker: Ticker) => void): () => void {
    const sub: Subscription = { symbol: symbol.toUpperCase(), callback };
    this.subscriptions = [...this.subscriptions, sub];  // ← O(n) copy on every subscribe
    // ...
}
```

## Growth Pattern
Every call to `watchTicker()` (via `BaseDataFeed.watchTicker()` → `watchTickerImpl()`) creates a brand-new array containing all existing subscriptions plus the new one. Every unsubscribe call similarly creates a new filtered array.

This is O(n) work per subscribe/unsubscribe and O(n²) total work to add n subscriptions:
- Subscription #1: copy 0 elements → 1-element array
- Subscription #2: copy 1 element → 2-element array
- ...
- Subscription #N: copy N-1 elements → N-element array
- Total allocations: 0 + 1 + 2 + ... + (N-1) = **N(N-1)/2 = O(N²)**

The WebSocket handler (`ws-handler.ts`) creates a new feed subscription for each WS client subscribing to `watchTicker`. In a server with many concurrent clients:

- 1000 concurrent clients → 1000 subscriptions
- Each subscribe creates and discards 999-element intermediate arrays
- Total GC load: ~500,000 elements worth of short-lived arrays

## OOM Estimate
The steady-state array at N subscriptions holds N `Subscription` objects (~100 bytes each):
- 1000 subscribers: array = 100 KB (manageable)
- OOM risk is not in the array itself but in the GC churn from O(N²) intermediate allocations during a burst of subscriptions
- At 100 subscribe/s with 1000 existing subscribers: 100 × 1000 × 100 bytes = **10 MB/s** of short-lived array garbage
- V8 minor GC triggered continuously, causing consistent streaming latency spikes

## Suggested Fix
Use a `Set` or an array with `push`/`splice` instead of the immutable spread pattern:

```typescript
// Replace the array field:
private subscriptions: Set<Subscription> = new Set();

// Subscribe:
this.subscriptions.add(sub);
return () => { this.subscriptions.delete(sub); };

// Dispatch (in handleMessage):
for (const sub of this.subscriptions) {
    if (sub.symbol === ticker.symbol) sub.callback(ticker);
}
```

`Set.add()` and `Set.delete()` are O(1) amortized, eliminating both the O(N²) subscription cost and the per-unsubscribe O(N) filter.

Apply to both `BinanceFeed` and `ChainlinkFeed`.

---

### LINKED ISSUE #719: Unhandled async: gemini-titan/websocket.ts — connect() fire-and-forget logs error but never calls scheduleReconnect on synchronous failure
## Risk Level
MEDIUM

## Location
`core/src/exchanges/gemini-titan/websocket.ts:268–271` (`watchOrderBook`) and `302–305` (`watchTrades`)

## Code
Both `watchOrderBook` (line 268) and `watchTrades` (line 302) use the same fire-and-forget pattern:

```typescript
if (!this.isConnected) {
    this.connect().catch((err: unknown) => {
        logger.warn(`[gemini-titan] connect failed during watchOrderBook('${symbol}'): ${err instanceof Error ? err.message : String(err)}`);
    });
}
```

The catch handler logs the error — which is better than the Kalshi (#678) and Opinion (#679) cases — but it does **not** call `scheduleReconnect()`. This means a synchronous failure inside `connect()` permanently kills the reconnect loop.

The relevant section of `connect()`:

```typescript
this.connectionPromise = new Promise((resolve, reject) => {
    try {
        const headers: Record<string, string> = this.auth
            ? this.auth.buildWsHeaders()   // ← can throw synchronously
            : {};

        this.ws = new WebSocket(this.config.wsUrl, { headers, handshakeTimeout: 30_000 });

        this.ws.on('close', () => {
            // scheduleReconnect() only reachable if WebSocket was created
            if (!this.isTerminated) {
                this.scheduleReconnect();
            }
        });
    } catch (error) {
        this.isConnecting = false;
        this.connectionPromise = undefined;
        reject(error);   // ← scheduleReconnect() is NEVER called here
    }
});
```

## What Happens on Failure
Concrete trigger: `GeminiAuth.buildWsHeaders()` throws synchronously (e.g., invalid or expired signing key, malformed private key).

When this happens:
1. `connect()` rejects before any WebSocket object is created.
2. The `'close'` event never fires → `scheduleReconnect()` is never queued.
3. The `.catch(err => logger.warn(...))` swallows the rejection after logging.
4. No future reconnection is ever attempted.
5. `dataPromise` (the resolver waiting for the first orderbook/trade event) hangs until `withWatchTimeout` fires (default 30 s).
6. **With `watchTimeoutMs: 0`, the resolver hangs indefinitely.**

The same structural break affects the `scheduleReconnect()` → `connect()` path (line 127–131): if a scheduled reconnect also fails synchronously, the loop is again not rescheduled.

## Suggested Fix
Call `scheduleReconnect()` in the `.catch()` handler so the reconnect loop stays alive regardless of how `connect()` failed:

```typescript
this.connect().catch((err: unknown) => {
    logger.warn(`[gemini-titan] connect failed during watchOrderBook('${symbol}')`, {
    });
    // Synchronous failure — 'close' event never fires, so schedule reconnect manually.
    this.scheduleReconnect();
});
```

Apply the same fix to the `watchTrades` call site (line 302) and to the `scheduleReconnect()` → `connect()` `.catch()` at line 128.

---

### LINKED ISSUE #727: Unbounded: core/src/exchanges/opinion/websocket.ts — applyLevelUpdate allocates new array + handleDepthDiff copies OrderBook object on every WebSocket message
## Location
`core/src/exchanges/opinion/websocket.ts:246-258` (`handleDepthDiff`)
`core/src/exchanges/opinion/websocket.ts:462-483` (`applyLevelUpdate`)

## Code
```typescript
// handleDepthDiff — line 245
private handleDepthDiff(data: any): void {
    const existing = this.orderBooks.get(marketId);
    const book: OrderBook = existing
        ? { ...existing }                         // ← shallow copy of OrderBook on EVERY message
        : { bids: [], asks: [], timestamp: Date.now() };

    if (side === "bids") {
        book.bids = applyLevelUpdate(book.bids, price, size, "desc");  // ← new array
    } else if (side === "asks") {
        book.asks = applyLevelUpdate(book.asks, price, size, "asc");   // ← new array
    }

    book.timestamp = Date.now();
    this.orderBooks.set(marketId, book);
    this.resolveOrderBook(marketId, book);
}

// applyLevelUpdate — line 462
function applyLevelUpdate(
    levels: readonly OrderLevel[],
    price: number,
    size: number,
    sortOrder: "asc" | "desc",
): OrderLevel[] {
    const updated = levels.filter(             // ← creates new array of n-1 elements
        (l) => Math.abs(l.price - price) >= 0.0001,
    );

    if (!isNaN(size) && size > 0) {
        updated.push({ price, size });         // ← new { price, size } object
    }

    updated.sort((a, b) => ...);               // ← in-place sort on new array

    return updated;
}
```

## Growth Pattern
Every `market.depth.diff` WebSocket message triggers:
1. `{ ...existing }` — shallow copy of the `OrderBook` object (~3 properties)
2. `levels.filter(...)` — allocates a new `OrderLevel[]` array of N elements (where N = current order book depth, typically 20–200 levels per side)
3. `{ price, size }` — one new `OrderLevel` object if inserting a new level
4. Old `OrderBook` object and old levels array become garbage immediately after `this.orderBooks.set(marketId, book)`

Compare with `KalshiWebSocket.applyDelta` and `GeminiWebSocket.applyDelta`, which mutate the levels array **in place** (using `splice` and `push`), avoiding any allocation on the hot path.

`resolveOrderBook` then copies both arrays again via `[...orderBook.bids]` / `[...orderBook.asks]` (line 265–266) before dispatching to resolvers, doubling the per-message allocation.

## OOM Estimate
Opinion Trade is a live binary prediction market. Active markets can have 50–200 depth diff messages per second during trading:

| Active markets | Msg rate | Levels per side | Allocations per msg | Memory churn |
|---|---|---|---|---|
| 10 | 20 msg/s | 50 levels | 2 arrays × 50 = 100 elems + 2 obj copies | ~400 KB/s garbage |
| 50 | 50 msg/s | 100 levels | 2 arrays × 100 = 200 elems + 2 obj copies | ~5 MB/s garbage |
| 200 | 100 msg/s | 100 levels | 2 arrays × 100 = 200 elems + 2 obj copies | ~40 MB/s garbage |

At 200 subscribed markets under load: V8 minor GC fires every few hundred ms, causing consistent latency jitter in the streaming callbacks that dispatch order book snapshots to WebSocket clients.

The `resolveOrderBook` double-copy (lines 264–267) runs at the same rate and doubles the pressure.

## Suggested Fix
**1. Mutate the order book in place** (matching the Kalshi/Gemini pattern):

```typescript
private handleDepthDiff(data: any): void {
    const marketId: number = data.marketId;
    const price = parseFloat(data.price);
    const size = parseFloat(data.size);
    const side: string = data.side;

    let book = this.orderBooks.get(marketId);
    if (!book) {
        book = { bids: [], asks: [], timestamp: Date.now() };
        this.orderBooks.set(marketId, book);
    }

    const levels = side === "bids" ? book.bids : book.asks;
    const sortOrder = side === "bids" ? "desc" : "asc";
    applyLevelUpdateInPlace(levels, price, size, sortOrder);  // mutate, no new array

    book.timestamp = Date.now();
    this.resolveOrderBook(marketId, book);
}
```

**2. Apply delta in-place** (no filter/new array):

```typescript
function applyLevelUpdateInPlace(
    levels: OrderLevel[],
    price: number,
    size: number,
    sortOrder: "asc" | "desc",
): void {
    const idx = levels.findIndex(l => Math.abs(l.price - price) < 0.0001);
    if (isNaN(size) || size <= 0) {
        if (idx !== -1) levels.splice(idx, 1);
    } else if (idx !== -1) {
        levels[idx].size = size;
    } else {
        levels.push({ price, size });
        levels.sort((a, b) => sortOrder === "desc" ? b.price - a.price : a.price - b.price);
    }
}
```

**3. Remove the double-copy in `resolveOrderBook`** — callers that need a snapshot can copy lazily.

---

### LINKED ISSUE #809: empty-catch: core/src/feeds/chainlink/chainlink-feed.ts:387 — JSON parse failure silently swallowed
## Violation
```ts
try {
    msg = JSON.parse(text) as ChainlinkWsMessage;
} catch {
    return;
}
```

## Location
`core/src/feeds/chainlink/chainlink-feed.ts:387`

## Why It Matters
A malformed or unexpected WebSocket frame from Chainlink causes `handleMessage` to silently return. There is no log at any level, so recurring parse failures (e.g. a Chainlink protocol change or a partial message) are completely invisible in production. Operators have no way to know the feed is receiving junk.

## Suggested Fix
```ts
} catch (err) {
    logger.warn('ChainlinkFeed: failed to parse WS message', { error: String(err) });
    return;
}
```

---

### LINKED ISSUE #810: empty-catch: core/src/feeds/binance/binance-feed.ts:216 — JSON parse failure silently swallowed
## Violation
```ts
try {
    msg = JSON.parse(text) as BinanceRelayMessage;
} catch {
    return;
}
```

## Location
`core/src/feeds/binance/binance-feed.ts:216`

## Why It Matters
A malformed frame from the Binance relay WebSocket causes `handleMessage` to silently return with no log. Recurring parse failures (e.g. relay protocol change, partial frame) will be completely invisible in production, making it impossible to diagnose a broken feed.

## Suggested Fix
```ts
} catch (err) {
    logger.warn('BinanceFeed: failed to parse relay WS message', { error: String(err) });
    return;
}
```

---

### LINKED ISSUE #811: empty-catch: core/src/exchanges/gemini-titan/websocket.ts:94 — unparseable messages silently ignored
## Violation
```ts
try {
    const message = JSON.parse(data.toString());
    this.handleMessage(message);
} catch {
    // Ignore unparseable messages
}
```

## Location
`core/src/exchanges/gemini-titan/websocket.ts:94`

## Why It Matters
The catch block contains only a comment and performs no action. Any JSON parse error or runtime exception thrown inside `handleMessage` is completely swallowed. This masks both protocol mismatches and bugs in the message handler itself. A warn-level log would allow operators to detect when Gemini Titan starts sending unexpected frames.

## Suggested Fix
```ts
} catch (err) {
    logger.warn('GeminiTitanWs: failed to parse or handle message', { error: String(err) });
}
```

---

### LINKED ISSUE #812: empty-catch: core/src/exchanges/polymarket_us/websocket.ts:127 — socket close errors silently ignored
## Violation
```ts
try {
    this.socket.close();
} catch {
    // Ignore close errors.
}
```

## Location
`core/src/exchanges/polymarket_us/websocket.ts:127`

## Why It Matters
The catch block contains only a comment and performs no action. While ignoring `close()` errors is often intentional, swallowing them completely hides any unexpected exception (e.g. a torn-down socket throwing a non-close-related error). A debug-level log costs nothing and makes tear-down diagnostics easier.

## Suggested Fix
```ts
} catch (err) {
    logger.debug('PolymarketUSWs: error during socket close', { error: String(err) });
}
```

---

realfishsam:
## Summary
- Log previously swallowed websocket close/parse/connect errors across Gemini Titan, Kalshi, Opinion, and Polymarket US
- Avoid per-message array/object allocations in Opinion and Binance/Chainlink feed paths
- Make Gemini Titan orderbook level matching exact instead of tolerance-based

## Validation
- git diff --check

Fixes #811
Fixes #719
Fixes #275
Fixes #678
Fixes #679
Fixes #727
Fixes #809
Fixes #810
Fixes #812
Fixes #686
Fixes #687


realfishsam:
## PR Review: FAIL

### What This Does
Surfaces websocket/feed hygiene failures more explicitly instead of silently masking unsupported or malformed feed/websocket states.

### Blast Radius
Several exchange websocket adapters plus Binance/Chainlink feed code and SDK API references.
Changed files reviewed: core/api-doc-config.generated.json, core/src/exchanges/gemini-titan/websocket.ts, core/src/exchanges/kalshi/websocket.ts, core/src/exchanges/opinion/websocket.ts, core/src/exchanges/polymarket_us/websocket.ts, core/src/feeds/binance/binance-feed.ts, core/src/feeds/chainlink/chainlink-feed.ts, sdks/python/API_REFERENCE.md, sdks/typescript/API_REFERENCE.md

### Consumer Verification
**Before (base branch):**

**After (PR branch):**

### Test Results
- Server starts: PASS

### Findings
1. Generated/API-doc artifacts changed without an obvious source/generator change; verify these are regenerated outputs, not hand edits. Files: core/api-doc-config.generated.json

### PMXT Pipeline Check
- OpenAPI sync: N/A
- Financial precision: OK
- Type safety: OK
- Auth safety: N/A

### Semver Impact

### Risk

PR #877, opened 2026-06-05
fix: reject throttler queue overflow

### LINKED ISSUE #690: Unbounded: core/src/utils/throttler.ts — queue overflow silently resolves dropped requests, bypassing rate limiting
## Location
`core/src/utils/throttler.ts:24-38` (`throttle()`)

## Code
```typescript
async throttle(cost: number = 1): Promise<void> {
    return new Promise<void>((resolve) => {
        if (this.queue.length >= this.maxQueueDepth) {
            const dropped = this.queue.shift();  // ← evict OLDEST waiter
            if (dropped) {
                dropped.resolve();  // ← silently resolve without rate-limit delay
            }
        }
        this.queue.push({ resolve, cost });
        if (!this.running) {
            this.running = true;
            this.loop();
        }
    });
}
```

When the queue is full (`maxQueueDepth`, default 1000), the **oldest** pending request is resolved immediately — as if rate-limiting completed — and then the **new** request is enqueued. The dropped caller proceeds to make its API call without having waited the required interval.

## Growth Pattern
This is not an OOM issue (the queue is bounded at 1000 by default), but it represents incorrect backpressure under overload:

1. **Rate-limit bypass**: Under sustained overload, the oldest callers are released without waiting. They issue their requests simultaneously, causing a burst of API calls that exceeds the configured rate limit. Exchanges respond with 429 errors.
2. **LIFO-like starvation**: The oldest requests (earliest in queue) are always the ones dropped, creating LIFO-like behavior where long-waiting requests never complete successfully.
3. **Silent failure**: The dropped `resolve()` returns `void` with no indication of being dropped. Callers cannot distinguish between "throttled successfully" and "dropped from queue" — they proceed as if throttling passed.

## OOM Estimate
N/A — queue is bounded. This is a correctness/safety issue that manifests as unexpected 429 bursts and silent request drops under load exceeding 1000 concurrent throttled calls.

## Suggested Fix
Replace the silent `dropped.resolve()` with `dropped.reject(new Error('Throttler queue full — request dropped'))` so callers can handle overload:

```typescript
if (this.queue.length >= this.maxQueueDepth) {
    const dropped = this.queue.shift();
    if (dropped) {
        // Reject so the caller knows the request was dropped, not rate-limited
        dropped.reject(new Error('Throttler queue capacity exceeded'));
    }
}
```

Alternatively, reject the **incoming** request instead of evicting the oldest:
```typescript
if (this.queue.length >= this.maxQueueDepth) {
    return Promise.reject(new Error('Throttler queue capacity exceeded'));
}
```

---

realfishsam:
## Summary
- Reject new throttle requests once the queue reaches its maximum depth instead of silently resolving the oldest waiter.

## Tests
- npm test --workspace=pmxt-core -- --runInBand test/utils/throttler.test.ts

Fixes #690


realfishsam:

### What This Does
Makes throttler queue overflow reject callers immediately instead of leaving pending promises. Consumers benefit on any rate-limited path under queue pressure.

### Blast Radius
Shared core throttler utility and tests; all exchange/server code using the throttler may be affected.
Changed files reviewed: core/src/utils/throttler.ts, core/test/utils/throttler.test.ts

### Consumer Verification
**Before (base branch):**

**After (PR branch):**

### Test Results
- Server starts: PASS

### Findings
No blocking findings.

### PMXT Pipeline Check
- OpenAPI sync: N/A
- Financial precision: N/A
- Type safety: ISSUE
- Auth safety: N/A

### Semver Impact

### Risk

PR #951, opened 2026-06-08
fix(python): guard nullable market slug cache

### LINKED ISSUE #156: Python load_markets: no None-guard on market.slug causes KeyError/TypeError
## Inconsistency

The Python SDK's `load_markets()` assigns `self.markets_by_slug[market.slug] = market` without checking whether `market.slug` is `None`, causing a `TypeError` (unhashable type `NoneType`) or `KeyError` at runtime. The TypeScript SDK correctly guards this assignment.

## Current (PMXT)

`sdks/python/pmxt/client.py` — `load_markets()` method:
```python
self.markets_by_slug[market.slug] = market  # slug can be None → TypeError
```

TypeScript SDK correctly guards this:
```typescript
// core/src/BaseExchange.ts (TypeScript)
if (market.slug) {
    this.marketsBySlug[market.slug] = market;
}
```

## Expected (CCXT convention)

CCXT's `loadMarkets` skips index entries when the key is falsy. The Python SDK should mirror the TypeScript guard:

```python
if market.slug:
    self.markets_by_slug[market.slug] = market
```

## Impact

Any venue that does not set `slug` on returned markets (e.g. Kalshi markets) will cause `load_markets()` to raise a `TypeError` in Python, making the entire exchange instance unusable. This is a runtime crash, not just a type mismatch.

realfishsam:
## Summary
- Skip slug index entries when a loaded market has no slug
- Mirrors the TypeScript loadMarkets guard and prevents crashes on slugless venues

Fixes #156

## Test Plan
- [x] git diff --check
- [x] python3 -m py_compile sdks/python/pmxt/client.py

### DIFF
--- a/sdks/python/pmxt/client.py
+++ b/sdks/python/pmxt/client.py
@@ -763,7 +763,8 @@ class Exchange(ABC):
 
         for market in markets:
             self.markets[market.market_id] = market
-            self.markets_by_slug[market.slug] = market
+            if market.slug:
+                self.markets_by_slug[market.slug] = market
 
         self._loaded_markets = True
         return self.markets


realfishsam:

### What This Does
Guards the Python SDK market slug cache so markets with `slug=None` are still cached by market id but no longer create a `None` key in `markets_by_slug`.

### Blast Radius
Python SDK only: `Exchange.load_markets()` cache population. No core/server/OpenAPI changes.

### Consumer Verification
**Before (base branch):**
On base, a `UnifiedMarket` with `slug=None` would execute `self.markets_by_slug[market.slug] = market`, creating a `None` lookup entry. I did not run a Python consumer repro because this runner is missing pytest for the SDK test phase.

**After (PR branch):**
Patch adds `if market.slug:` before writing the slug cache. Core build passed and core Jest tests passed; full root verification stopped in the Python SDK phase with `/usr/bin/python3: No module named pytest`.

### Test Results
- Build: PASS
- Unit tests: CORE PASS (24 suites, 644 tests; root verification then blocked by missing pytest)
- E2E smoke: NOT VERIFIED (Python SDK environment missing pytest)

### Findings
No blocking findings.

### PMXT Pipeline Check
- OpenAPI sync: N/A
- Financial precision: N/A
- Type safety: OK
- Auth safety: N/A

### Semver Impact
patch -- Python SDK cache bug fix without API shape change.

### Risk
Python SDK runtime behavior was inspected from the diff but not executed end-to-end in this environment because pytest is not installed.

PR #956, opened 2026-06-08
fix: align SDK parity gaps

### LINKED ISSUE #656: SDK drift: `Router.fetch_market_matches` calls server endpoint `"fetchMatches"` in Python but `"fetchMarketMatches"` in TypeScript
## SDK Drift Report

**Category:** Method behaviour drift (wrong server endpoint)

**TypeScript location:** `sdks/typescript/pmxt/router.ts` line 255 — calls `'fetchMarketMatches'`

**Python location:** `sdks/python/pmxt/router.py` line 201 — calls `"fetchMatches"` (stale endpoint name)

---

### What the drift is

The TypeScript `Router.fetchMarketMatches` dispatches to the correct sidecar endpoint:

```ts
// sdks/typescript/pmxt/router.ts line 255
return this.callMethod('fetchMarketMatches', params);
```

The Python `Router.fetch_market_matches` dispatches to a **different, apparently stale** endpoint name:

```python
# sdks/python/pmxt/router.py line 201
return await self._call_method("fetchMatches", params)
```

`"fetchMatches"` was the old name before the endpoint was renamed to `"fetchMarketMatches"`. The Python router was not updated.

---

### Why it matters

Python callers invoking `router.fetch_market_matches(...)` will hit the wrong (or non-existent) server endpoint, receiving either an error or stale/incorrect results while the TypeScript counterpart works correctly.

---

### Expected behaviour

`Router.fetch_market_matches` in Python should call `"fetchMarketMatches"` to match the TypeScript implementation.

### LINKED ISSUE #701: [sdk-drift] Router auto_start_server defaults differ: TypeScript inherits isHosted logic, Python always defaults to False
### Summary

The TypeScript `Router` inherits `Exchange`'s `autoStartServer` logic: it defaults to `true` in local (non-hosted) mode and `false` in hosted mode. The Python `Router` hard-codes `auto_start_server=False` regardless of mode, so it never auto-starts the sidecar.

---

### TypeScript

**File**: `sdks/typescript/pmxt/client.ts`, lines 266–271 (in `Exchange` constructor, inherited by `Router`)

```typescript
const autoStartServer = options.autoStartServer !== undefined
    ? options.autoStartServer
    : !this.isHosted;
// When no baseUrl is supplied and no pmxtApiKey, isHosted = false,
// so autoStartServer = true — the sidecar is started automatically.
```

`Router` is defined in `sdks/typescript/pmxt/router.ts` and passes options directly to `super("router", options)`, so it inherits this behavior unchanged.

---

### Python

**File**: `sdks/python/pmxt/router.py`, lines 72–91

```python
class Router(Exchange):
    def __init__(
        self,
        pmxt_api_key: Optional[str] = None,
        base_url: Optional[str] = None,
        auto_start_server: bool = False,   # hardcoded False
        **kwargs,
    ):
        super().__init__(
            "router",
            auto_start_server=auto_start_server,
            ...
        )
```

The default `False` means a Python `Router()` constructed without a `pmxtApiKey` and without an explicit `base_url` will point at `http://localhost:3847` but will **not** attempt to start the sidecar. The TypeScript equivalent would auto-start the sidecar in the same configuration.

---

### Impact

- A Python user following the TypeScript "just instantiate `Router()`" pattern will get `ConnectionRefusedError` (or equivalent) if the sidecar is not already running, while the TypeScript user would have had the sidecar started for them.
- The behavioral difference is invisible from the constructor signatures — both accept an `auto_start_server` / `autoStartServer` override, but the defaults diverge.
- This is a behavioral drift, not just a type drift.

### LINKED ISSUE #702: [sdk-drift] stopServer/restartServer are permanent first-class API in TypeScript but deprecated in Python
### Summary

`pmxt.stopServer()` and `pmxt.restartServer()` are documented as permanent, fully-supported top-level aliases in TypeScript. The equivalent `pmxt.stop_server()` and `pmxt.restart_server()` in Python emit a `DeprecationWarning` directing callers to use `pmxt.server.stop()` and `pmxt.server.restart()` instead.

---

### TypeScript

**File**: `sdks/typescript/index.ts`, lines 39–47

```typescript
// Flat aliases for the namespaced server commands. Kept as permanent,
// fully-supported shorthand — `pmxt.server.stop()` and `pmxt.stopServer()`
// are equivalent and both are first-class API.
export const stopServer = server.stop.bind(server);
export const restartServer = server.restart.bind(server);
```

The comment is explicit: these are not deprecated; they are "permanent, fully-supported shorthand."

---

### Python

**File**: `sdks/python/pmxt/__init__.py`, lines 111–131

```python
def stop_server():
    warnings.warn(
        "stop_server() is deprecated, use pmxt.server.stop() instead",
        DeprecationWarning,
        stacklevel=2,
    )
    return server.stop()

def restart_server():
    warnings.warn(
        "restart_server() is deprecated, use pmxt.server.restart() instead",
        DeprecationWarning,
        stacklevel=2,
    )
    return server.restart()
```

---

### Impact

- Code written against the TypeScript SDK using `pmxt.stopServer()` / `pmxt.restartServer()` is correct and forward-compatible.
- Equivalent code written against the Python SDK using `pmxt.stop_server()` / `pmxt.restart_server()` produces deprecation warnings and may break in a future release.
- Cross-language documentation, examples, and tutorials that show the flat alias pattern will generate warnings for Python users.
- The API stability contract is inconsistent: TypeScript users can rely on these aliases; Python users cannot.

### LINKED ISSUE #745: Trade type missing outcomeId field in both SDKs
## Gap
The core `Trade` type includes `outcomeId?: string` — the outcome this trade belongs to. Neither the TypeScript SDK `Trade` interface nor the Python SDK `Trade` dataclass includes this field. When the sidecar returns trade objects with `outcomeId` set, the field is silently dropped by the SDK mapping layer.

## Core
**File:** `core/src/types.ts` line 159
```ts
export interface Trade {
    id: string;
    timestamp: number;
    price: number;
    amount: number;
    side: 'buy' | 'sell' | 'unknown';
    /** The outcome this trade is for (if known). */
    outcomeId?: string;
}
```
`outcomeId` identifies which outcome token the trade was for. This is populated by exchanges that return per-outcome trade history — for example, Polymarket trades on a specific CLOB token ID. Without it, callers who receive a flat list of trades across outcomes cannot re-group them.

Note: `UserTrade extends Trade`, so `UserTrade` also inherits this field in the core. The Python SDK `UserTrade` explicitly adds `outcome_id` (models.py line 337), but the base `Trade` type does not have it, creating an inconsistency.

## TypeScript SDK
Missing — `sdks/typescript/pmxt/models.ts` `Trade` interface has five fields (`id`, `timestamp`, `price`, `amount`, `side`) but no `outcomeId`. The `UserTrade` interface (models.ts line 199) adds `outcomeId?: string` separately — but the base `Trade` type, which is what `fetchTrades()` returns, lacks it.

## Python SDK
Missing from base — `sdks/python/pmxt/models.py` `Trade` dataclass (line 310) has five fields (`id`, `timestamp`, `price`, `amount`, `side`) but no `outcome_id`. The `UserTrade` dataclass (line 329) adds `outcome_id: Optional[str]` explicitly, but the base `Trade` class, which is what `fetch_trades()` returns, does not.

## Evidence
`grep -n "outcomeId" sdks/typescript/pmxt/models.ts` returns line 219 (in `UserTrade` only, not in `Trade`).
`grep -n "outcome_id" sdks/python/pmxt/models.py` returns line 337 (in `UserTrade` only, not in `Trade`).
`grep -n "outcomeId" core/src/types.ts` returns line 159 (in `Trade`).
Both SDKs' `_convert_trade` / `convertTrade` internal helpers map to their respective `Trade` types — if the field is absent from the type, it is stripped from the object even if the sidecar populates it.

## Impact
Callers using `fetchTrades(outcomeId)` receive trades already scoped to an outcome, so they may not need the field. But callers who use exchange-level trade streams (or reconstruct trade history across outcomes) cannot identify which outcome each `Trade` belongs to. The asymmetry with `UserTrade` — which does carry `outcomeId` — is confusing: the same underlying data is only partially typed depending on which method returned the trade.

---
*Found by automated Core-to-SDK surface coverage audit*

### LINKED ISSUE #929: sdk-drift: ExchangeOptions.signatureType typed as number-only in TypeScript; Python accepts any value including strings
## Drift

`signatureType` / `signature_type` accepts different value domains across the two SDKs. TypeScript restricts it to `number` for the base `ExchangeOptions` type, making string values like `"gnosis-safe"` a type error for all non-Polymarket exchanges. Python accepts any value for all exchanges.

## TypeScript SDK

`sdks/typescript/pmxt/models.ts` — `ExchangeOptions` interface:
```ts
signatureType?: number;
```
Only `PolymarketOptions` (a subtype) widens this to `'eoa' | 'poly-proxy' | 'gnosis-safe' | number`. Every other exchange inherits the base `number`-only constraint.

## Python SDK

`sdks/python/pmxt/client.py` — `Exchange.__init__` (line ~295):
```python
signature_type: Optional[Any] = None
```
All venue subclasses in `sdks/python/pmxt/_exchanges.py` inherit this; no narrowing is applied for any venue including Polymarket.

## Expected

Both SDKs should expose the same accepted value domain for `signatureType` / `signature_type`. Either:
- TypeScript widens `ExchangeOptions.signatureType` to `string | number` (or the same union as `PolymarketOptions`), or
- Python narrows the type per venue to match TypeScript's intent.

## Impact

TypeScript users passing `signatureType: "gnosis-safe"` (or any string) to Kalshi, Limitless, or any exchange other than Polymarket receive a compile-time type error. Python users pass the same string without issue. Shared configuration objects and documentation examples that use string values will silently diverge between SDKs.

---

### LINKED ISSUE #930: sdk-drift: create_order/build_order validation errors raise ValueError in Python but PmxtError in TypeScript
## Drift

Input-validation failures inside `createOrder` / `create_order` and `buildOrder` / `build_order` raise different exception types in the two SDKs. TypeScript raises `PmxtError`; Python raises the built-in `ValueError`.

## TypeScript SDK

`sdks/typescript/pmxt/client.ts` (lines ~1955, ~2105):
```ts
throw new PmxtError("Cannot specify both 'outcome' and 'marketId'/'outcomeId'...");
```
All validation branches inside `buildOrder` and `createOrder` throw `PmxtError`, which is the SDK's own error hierarchy.

## Python SDK

`sdks/python/pmxt/client.py` (lines ~2427, ~2547):
```python
raise ValueError("Cannot specify both 'outcome' and 'market_id'/'outcome_id'...")
```
The same validation branches raise Python's built-in `ValueError` instead of a `PmxtError` subclass.

## Expected

Validation failures should raise a `PmxtError` subclass in both SDKs. The appropriate subclass is `ValidationError` (already defined in both `sdks/typescript/pmxt/errors.ts` and `sdks/python/pmxt/errors.py`), which carries an optional `field` / `field` attribute.

## Impact

Users who catch `PmxtError` (or `pmxt.ValidationError`) in a try/except block to handle bad-input cases will silently miss validation errors in the Python SDK, because `ValueError` is not a subclass of `PmxtError`. Code ported between SDKs requires different exception-handling logic for identical error conditions.

---

### LINKED ISSUE #935: Python SDK `_get_credentials_dict` does not forward `walletAddress` — Limitless delegated signing broken
## Gap

The core `ExchangeCredentials` interface includes `walletAddress?: string` — used by `LimitlessExchange` for delegated signing (looking up a profile ID from a wallet address). The Python SDK's `_get_credentials_dict` method, which serializes credentials into the sidecar request body, does not include `walletAddress` in the dict it builds. Even if a caller somehow passes `wallet_address` during construction (it is not a constructor parameter on the base `Exchange`), it would be silently dropped before reaching the sidecar.

## Core

**`core/src/BaseExchange.ts:371-392`**
```typescript
export interface ExchangeCredentials {
    apiKey?: string;
    apiSecret?: string;
    passphrase?: string;
    apiToken?: string;
    privateKey?: string;
    signatureType?: number | string;
    funderAddress?: string;
    walletAddress?: string;   // ← used by Limitless for delegated signing
    baseUrl?: string;
}
```

**`core/src/exchanges/limitless/index.ts:104,115`** — `walletAddress` is read from `credentials.walletAddress` and used to look up the signing profile for delegated orders.

## Python SDK

**`sdks/python/pmxt/client.py:490-506`**
```python
def _get_credentials_dict(self) -> Optional[Dict[str, Any]]:
    """Build credentials dictionary for API requests."""
    if not self.api_key and not self.private_key and not self.api_token:
        return None

    creds = {}
    if self.api_key:
        creds["apiKey"] = self.api_key
    if self.private_key:
        creds["privateKey"] = self.private_key
    if self.api_token:
        creds["apiToken"] = self.api_token
    if self.proxy_address:
        creds["funderAddress"] = self.proxy_address
    if self.signature_type is not None:
        creds["signatureType"] = self.signature_type
    return creds if creds else None
```

`walletAddress` is never set in this dict. The base `Exchange.__init__` also has no `wallet_address` parameter, so there is no way to supply the value at all.

## TypeScript SDK

The TypeScript SDK's `getCredentials()` method in `client.ts` similarly builds a credentials object from the constructor options. `walletAddress` is not part of `ExchangeOptions` in the TypeScript SDK either, so neither SDK forwards this credential.

Note: Issue #853 tracks the per-exchange `OpinionExchangeOptions.walletAddress` gap. This issue tracks the lower-level, cross-cutting failure: even if a future fix adds `walletAddress` to the exchange constructor, the Python credential serialization layer would still silently drop it.

## Evidence

`grep -n "walletAddress\|wallet_address" sdks/python/pmxt/client.py` → zero results.
`grep -n "walletAddress" core/src/BaseExchange.ts` → line 388.
`grep -n "walletAddress" core/src/exchanges/limitless/index.ts` → lines 104, 115 confirming the field is consumed by exchange logic.

## Impact

Limitless delegated signing workflows cannot pass the `walletAddress` credential through the Python SDK, even if the caller knows the wallet address. The credential is absent from the serialized dict sent to the sidecar, so the sidecar-side `LimitlessExchange` sees `credentials.walletAddress === undefined` and falls through to the default signing path. Delegated signing fails silently — no error, just wrong behavior (orders placed on the caller's own profile instead of the delegated profile).

---
*Found by automated Core-to-SDK surface coverage audit*

realfishsam:
## Summary
- forward Limitless wallet addresses through Python SDK credentials for delegated signing
- align Python order validation failures with the SDK ValidationError hierarchy
- add missing trade outcome ID fields to both SDK models
- call the current router fetchMarketMatches endpoint and align Python Router auto-start defaults
- widen TypeScript ExchangeOptions.signatureType to accept string values
- keep Python stop_server/restart_server as non-deprecated first-class aliases

Fixes #935
Fixes #930
Fixes #745
Fixes #656
Fixes #701
Fixes #929
Fixes #702

## Test Plan
- python3 -m compileall sdks/python/pmxt
- npm run generate:docs --workspace=pmxt-core (no substantive generated doc diffs)
- npm run build --workspace=pmxtjs (blocked: generated SDK missing, and regenerating requires java which is not installed in this environment)
- npm run build --workspace=pmxt-core (attempted; timed out in this environment)


realfishsam:

### What This Does
Aligns SDK parity gaps: Python server aliases, wallet-address credential plumbing, ValidationError usage, order/trade fields, and TypeScript signature type/outcome fields.

### Blast Radius
Python SDK exports/client/router/models and TypeScript client/models; no core server schema changes.

### Consumer Verification
**Before (base branch):**
Base SDKs had parity mismatches such as deprecated warnings for flat server aliases, missing walletAddress credential forwarding, and `fetchMatches` call naming drift.

**After (PR branch):**
PR updates those SDK surfaces. Core build/Jest passed (24 suites, 644 tests); root verification then failed in the Python SDK phase because pytest is not installed.

### Test Results
- Build: PASS
- Unit tests: CORE PASS (24 suites, 644 tests; root verification blocked by missing pytest)
- E2E smoke: NOT VERIFIED (SDK import/call smoke not run)

### Findings
No blocking findings.

### PMXT Pipeline Check
- OpenAPI sync: N/A
- Financial precision: N/A
- Type safety: OK
- Auth safety: OK — credential field is forwarded, not logged

### Semver Impact
patch -- SDK parity/typing fix.

### Risk
SDK consumer behavior was not executed end-to-end because the local Python SDK test dependency is missing.

realfishsam:
Merge-refresh note from the low-hanging PR automation:

- Merged current `origin/main` into this branch and resolved SDK conflicts in:
  - `sdks/python/pmxt/client.py`
  - `sdks/typescript/pmxt/client.ts`
- Local targeted checks after the conflict fix:
  - `npm run build --workspace=pmxt-core` ✅
  - `python3 -m compileall -q sdks/python/pmxt` ✅
  - `node --check sdks/typescript/scripts/generate-client-methods.js` ✅
  - `git diff --check` ✅
- I did **not** merge this PR: GitHub generated-sync checks are red after the refresh. Re-running the client/docs generators locally would remove existing hosted-mode SDK sections / create broad generated drift unrelated to this focused SDK-parity PR, so I left that drift uncommitted.

Current blocker: generated client/docs sync needs a safer generator update or coordinated generated-output refresh before this is merge-ready.

realfishsam:

### What This Does
Aligns several Python/TypeScript SDK parity gaps: Python Limitless accepts `wallet_address`, credentials include proxy/signature/wallet fields, Python router calls `fetchMarketMatches`, error type mapping uses `ValidationError`, and trade/signature model types are widened.

### Blast Radius
Python SDK client/router/models/exports and TypeScript SDK type declarations. Core sidecar behavior is unchanged.

### Consumer Verification
**Before (base branch):**
Python Limitless could not pass a delegated `wallet_address` through the convenience wrapper, `Exchange._get_credentials_dict()` returned `None` when only proxy/signature/wallet fields were provided, and Python router convenience `fetch_market_matches` used the older `fetchMatches` method name.

**After (PR branch):**
The wrapper forwards `wallet_address`, credential serialization includes `proxyAddress`, `signatureType`, and `walletAddress`, router uses `fetchMarketMatches`, and Python syntax checks pass for all changed Python files.

### Test Results
- Core unit tests: PASS (27 suites / 652 tests passed; 1 suite / 3 tests skipped)
- Python syntax: PASS (`py_compile` on changed Python files)
- Full `npm test`: FAIL in this checkout during Python SDK collection because `pmxt_internal` / `eth_account` are not available.
- Server starts: N/A (SDK-only behavior)
- E2E smoke: Not run through live hosted/local SDK due missing generated Python package

### Findings
No blocking findings.

### PMXT Pipeline Check
- OpenAPI sync: N/A
- Financial precision: N/A
- Type safety: OK (signature type widened consistently; optional field additions)
- Auth safety: OK (credentials are serialized, not logged)

### Semver Impact
patch -- SDK parity/bug-fix additions and error-class alignment.

### Risk
I could not execute the Python SDK consumer path in this checkout because the generated `pmxt_internal` package is missing, so verification is by diff, syntax, and core tests rather than live SDK calls.

PR #961, opened 2026-06-08
fix(core): handle shutdown lock cleanup rejection

### LINKED ISSUE #720: Unhandled async: server/index.ts — async shutdown handler on SIGTERM/SIGINT; rejection prevents process.exit(0)
## Risk Level
HIGH

## Location
`core/src/server/index.ts:63–71`

## Code
```typescript
const shutdown = async () => {
    logger.info('Shutting down gracefully...');
    server.close();
    await lockFile.remove();   // ← can reject
    process.exit(0);           // ← never reached if lockFile.remove() rejects
};

process.on('SIGTERM', shutdown);
process.on('SIGINT', shutdown);
```

`shutdown` is an `async` function. Node.js process signal handlers receive the return value of the callback and ignore it — the returned Promise is **never awaited and has no `.catch()`**. If `lockFile.remove()` rejects, the rejection goes unhandled.

## What Happens on Failure
`lockFile.remove()` can fail if:
- The lock file was already deleted by another process (e.g., zombie restart).
- A file-system permission error or a full disk.

When it fails:
1. `lockFile.remove()` rejects inside `shutdown`.
2. The rejection propagates out of `shutdown` as an unhandled promise rejection.
3. In Node.js ≥ 15 (default `--unhandled-rejections=throw`), this emits `UnhandledPromiseRejection` — which itself terminates the process with **exit code 1**, not the intended 0. In earlier Node.js versions it is silently swallowed.
4. `process.exit(0)` is **never called**.
5. The server's HTTP listener was already closed by `server.close()`, so the process is left in a limbo state: it neither accepts connections nor exits.
6. The next SIGTERM also goes nowhere — the handler runs the same broken path.
7. Operators must resort to `SIGKILL` to stop the process.

## Suggested Fix
Wrap the async body in a try/catch so the handler always calls `process.exit()`, regardless of whether the lock-file removal succeeded:

```typescript
const shutdown = async () => {
    logger.info('Shutting down gracefully...');
    server.close();
    try {
        await lockFile.remove();
    } catch (err) {
        logger.warn('Failed to remove lock file during shutdown', {
        });
    }
    process.exit(0);
};

process.on('SIGTERM', () => {
    shutdown().catch((err) => {
        logger.error('Unexpected error in shutdown handler', { error: String(err) });
        process.exit(1);
    });
});
process.on('SIGINT', () => {
    shutdown().catch((err) => {
        logger.error('Unexpected error in shutdown handler', { error: String(err) });
        process.exit(1);
    });
});
```

---

realfishsam:
## Summary
- make signal shutdown handler synchronous so Node does not ignore a returned rejecting Promise
- catch lock-file cleanup failures, log them, and still exit with the intended success code
- guard against duplicate SIGTERM/SIGINT shutdown execution

Fixes #720

## Test Plan
- npm run build --workspace=pmxt-core

### DIFF
--- a/core/src/server/index.ts
+++ b/core/src/server/index.ts
@@ -60,11 +60,20 @@ async function main() {
     logger.info(`Lock file created at ${lockFile.lockPath}`);
 
     // Graceful shutdown
-    const shutdown = async () => {
+    let shuttingDown = false;
+    const shutdown = () => {
+        if (shuttingDown) return;
+        shuttingDown = true;
+
         logger.info('Shutting down gracefully...');
         server.close();
-        await lockFile.remove();
-        process.exit(0);
+        void lockFile.remove()
+            .catch((error) => {
+                logger.warn('Failed to remove lock file during shutdown:', error);
+            })
+            .finally(() => {
+                process.exit(0);
+            });
     };
 
     process.on('SIGTERM', shutdown);


realfishsam:

### What This Does
Makes server shutdown lock-file cleanup rejection-safe and idempotent so signal handling still exits after cleanup failures.

### Blast Radius
Core server startup/shutdown path only.

### Consumer Verification
**Before (base branch):**
Base shutdown awaited `lockFile.remove()` directly before `process.exit(0)`, so a rejected cleanup could prevent clean process exit.

**After (PR branch):**
PR uses a `shuttingDown` guard and always exits in `.finally()` after logging cleanup errors. Core build/Jest passed (24 suites, 644 tests); root verification then hit missing pytest.

### Test Results
- Build: PASS
- Unit tests: CORE PASS (24 suites, 644 tests; root verification blocked by missing pytest)
- E2E smoke: NOT VERIFIED with signal-level process test

### Findings
No blocking findings.

### PMXT Pipeline Check
- OpenAPI sync: N/A
- Financial precision: N/A
- Type safety: OK
- Auth safety: N/A

### Semver Impact
patch -- server shutdown robustness fix.

### Risk
Actual SIGINT/SIGTERM behavior with a failing lock-file removal was not reproduced; conclusion is from code path and passing build/tests.

PR #981, opened 2026-06-09
fix(core): batch Polymarket position condition_ids gamma lookup

atomzakk:
Chunk Gamma /markets enrichment in fetchPositions so wallets with 50+ historical positions no longer hit Cloudflare 403 HTML

### DIFF
--- a/core/src/exchanges/polymarket/fetcher.ts
+++ b/core/src/exchanges/polymarket/fetcher.ts
@@ -112,6 +112,9 @@ export interface PolymarketRawPosition {
 
 const GAMMA_MARKETS_URL = process.env.POLYMARKET_GAMMA_MARKETS_URL || 'https://gamma-api.polymarket.com/markets';
 
+// Cloudflare blocks large single-shot Gamma lookups (~50 condition_ids / ~3.5KB query).
+const GAMMA_CONDITION_IDS_BATCH_SIZE = 40;
+
 export class PolymarketFetcher implements IExchangeFetcher<PolymarketRawEvent, PolymarketRawEvent> {
     private readonly ctx: FetcherContext;
     private readonly http: AxiosInstance;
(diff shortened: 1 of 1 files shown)


realfishsam:
## PR Review: VERIFIED

### What This Does
This changes Polymarket `fetchPositions` market-id enrichment to split Gamma `/markets?condition_ids=...` lookups into batches of 40. SDK consumers with wallets containing many historical positions should no longer hit a single oversized Gamma query that can be blocked by Cloudflare.

### Blast Radius
Polymarket core only: `core/src/exchanges/polymarket/fetcher.ts`, specifically the private `fetchRawPositions` enrichment path from `conditionId` to Gamma `market.id`. No unified type, OpenAPI schema, SDK shim, router, or auth surface changed.

### Consumer Verification
**Before (base branch):**
I built `origin/main` and ran a consumer-path harness that calls `PolymarketFetcher.fetchRawPositions()` with 81 positions and a local Gamma `/markets` mock that returns 403 when a request carries more than 40 `condition_ids`.

```bash
node /opt/data/cron/tmp/pmxt-pr-review/verify-polymarket-batch.js
```

Response snippet:
```json
{"ok":false,"requestBatchSizes":[81],"error":"Request failed with status code 403","status":403}
```

**After (PR branch):**
The same harness against PR `a032e51645a20f826ecf8962e49ec66430a7d647` succeeded and resolved all 81 market ids via three Gamma requests.

```json
{"ok":true,"requestBatchSizes":[40,40,1],"count":81,"resolved":81,"sample":[{"conditionId":"cond-0","size":"1","avgPrice":"0.5","resolvedMarketId":"market-0"},{"conditionId":"cond-1","size":"1","avgPrice":"0.5","resolvedMarketId":"market-1"}]}
```

### Test Results
- Unit tests: PASS (`npm test --workspace=pmxt-core`: 24 passed, 1 skipped; 644 passed, 3 skipped)
- Server starts: PASS as part of root `npm test` verification script
- E2E smoke: PASS for the targeted Polymarket `fetchRawPositions` enrichment harness above; root `npm test` later failed in the Python SDK phase because `/usr/bin/python3: No module named pytest` in this environment, after core tests and server startup had already passed.

### Findings
No blocking findings.

### PMXT Pipeline Check
- Field propagation (3-layer): N/A — no unified fields added/removed.
- OpenAPI sync: N/A — no public method/schema signature changed.
- Financial precision: OK — no price/amount arithmetic changed.
- Type safety: OK — no new `any` beyond existing Gamma payload handling.
- Auth safety: N/A — no credential/signing path changed.

### Semver Impact
patch -- bug fix for Polymarket position enrichment reliability.

### Risk
The batch size is a fixed conservative constant (40) based on observed Cloudflare behavior; if Gamma later lowers its query-size threshold this could need another adjustment. Requests are sequential rather than parallel, so very large historical wallets may incur extra latency, but correctness improves and the affected path is private position enrichment.

PR #1045, opened 2026-06-12
fix(python-sdk): serialize approve amount_wei as string

### LINKED ISSUE #1025: SDK drift: `Escrow.approveTx` `amountWei` is typed `bigint` in TypeScript but `int` in Python
## Drift

The `amountWei` / `amount_wei` parameter of the escrow approval method uses incompatible numeric types: TypeScript uses JavaScript's native `bigint` while Python uses `int`. Although both are arbitrary-precision integers semantically, the TypeScript side also serializes the value as a string (`amountWei.toString()`) before sending to the server, while Python sends it as a JSON number. This creates a wire-format mismatch.

## TypeScript SDK

**`sdks/typescript/pmxt/escrow.ts:27`**

```typescript
async approveTx(token: string, amountWei?: bigint): Promise<unknown>
```

The value is serialized at line 34 as:
```typescript
amount_wei: amountWei === undefined ? null : amountWei.toString(),
```

Wire format: **string** (e.g. `"1000000000000000000"`).

## Python SDK

**`sdks/python/pmxt/escrow.py:46,114`**

```python
def _amount_wei(value: int | None) -> int | None: ...

def approve_tx(self, token: str, amount_wei: int | None = None) -> dict[str, Any]:
```

Wire format: **JSON number** (e.g. `1000000000000000000`).

## Expected

Both SDKs should accept the same input type and produce the same wire format. Large Wei values (e.g. `2^96 - 1` for unlimited approval) exceed JavaScript's safe integer range, so string serialization is correct. Python should also serialize `amount_wei` as a string, or the server must be documented to accept both.

## Impact

The server receives `amount_wei` as a string from TypeScript callers and as a JSON number from Python callers. If the server strictly validates the type, one of the two SDKs will fail. For large ERC-20 unlimited approvals, Python's integer JSON encoding may overflow the server's JSON parser if it uses a 64-bit integer type.

realfishsam:
## Summary
- Serialize Python `Escrow.approve_tx(..., amount_wei=...)` values as strings to match TypeScript `approveTx` wire format.
- Add a focused regression test for a large Wei approval amount.

Fixes #1025

## Test Plan
- `pytest sdks/python/tests/test_escrow_amount_wei.py::test_approval_amount_wei_serializes_as_string_for_json_wire_format -q`
- `pytest sdks/python/tests/test_escrow_amount_wei.py -q`

## Notes
- `sdks/python/tests/test_hosted_dispatch.py` cannot be collected directly in this checkout without generated `pmxt_internal`; the new helper-level test loads `pmxt.escrow` directly to avoid requiring generated SDK artifacts for this narrow regression.

### DIFF
--- a/sdks/python/pmxt/escrow.py
+++ b/sdks/python/pmxt/escrow.py
@@ -43,7 +43,7 @@ def _approval_token(token: str) -> str:
     )
 
 
-def _amount_wei(value: int | None) -> int | None:
+def _amount_wei(value: int | None) -> str | None:
     if value is None:
         return None
     if isinstance(value, bool) or not isinstance(value, int):
@@ -53,7 +53,7 @@ def _amount_wei(value: int | None) -> int | None:
         )
     if value < 0:
         raise ValidationError("amount_wei must be non-negative", field="amount_wei")
-    return value
+    return str(value)
 
 
 def _usdc_amount(value: float | Decimal, *, field: str = "amount") -> float:
--- a/sdks/python/tests/test_escrow_amount_wei.py
+++ b/sdks/python/tests/test_escrow_amount_wei.py
@@ -1,0 +1,36 @@
+"""Tests for hosted escrow transaction body helpers."""
+
+from __future__ import annotations
+
+import importlib.util
+import pathlib
+import sys
+import types
+
+
+def _load_escrow_module():
+    """Load pmxt.escrow without importing pmxt.__init__.
+
+    The generated pmxt_internal package is not checked in, and pmxt.__init__
+    imports the full client. These helper-level tests only need sibling pmxt
+    modules, so install a minimal package shell and load escrow directly.
+    """
+    package_root = pathlib.Path(__file__).resolve().parents[1] / "pmxt"
+    pkg = types.ModuleType("pmxt")
+    pkg.__path__ = [str(package_root)]
+    sys.modules.setdefault("pmxt", pkg)
+    spec = importlib.util.spec_from_file_location(
+        "pmxt.escrow",
+        package_root / "escrow.py",
+    )
+    assert spec is not None and spec.loader is not None
+    module = importlib.util.module_from_spec(spec)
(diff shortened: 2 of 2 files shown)


realfishsam:

### What This Does
Changes Python `Escrow.approve_tx(..., amount_wei=...)` helper serialization so large Wei values are sent as JSON strings instead of JSON numbers. This matters to hosted SDK consumers because integer Wei amounts can exceed JavaScript's safe integer range if they cross the HTTP boundary as numbers.

### Blast Radius
Python SDK escrow helper only: `sdks/python/pmxt/escrow.py` plus a focused helper-level regression test. No core sidecar/OpenAPI changes and no TypeScript SDK changes.

### Consumer Verification
**Before (base branch):**
Helper-level consumer path returned a Python integer:
```text
_call: _amount_wei(2**96 - 1)
result type/value: int 79228162514264337593543950335
```
That value would be JSON-encoded as a number by the request body builder.

**After (PR branch):**
The same helper-level call returns a string:
```text
_call: _amount_wei(2**96 - 1)
result type/value: str 79228162514264337593543950335
```
I did not run a live hosted escrow request; this was verified at the SDK request-body helper layer.

### Test Results
- Build: N/A for Python SDK helper change; `python3 -m py_compile sdks/python/pmxt/escrow.py sdks/python/tests/test_escrow_amount_wei.py` PASS
- Unit tests: NOT RUN (`python3` in this cron image has no `pytest` module)
- Server starts: N/A
- E2E smoke: NOT RUN (hosted escrow write path not exercised)

### Findings
No blocking findings.

### PMXT Pipeline Check
- OpenAPI sync: N/A
- Financial precision: OK — preserves exact Wei value by serializing as a string.
- Type safety: OK
- Auth safety: N/A

### Semver Impact
patch -- Python SDK wire-format bug fix for an existing helper.

### Risk
The actual HTTP request was not sent in this environment, and pytest was unavailable, so this is verified at helper/source level rather than full hosted-consumer E2E.

PR #1065, opened 2026-06-13
Forward TypeScript order optional params

### LINKED ISSUE #1031: TypeScript SDK `createOrder()` and `buildOrder()` silently drop `tickSize`, `negRisk`, and `onBehalfOf` from `CreateOrderParams`
## Gap

`CreateOrderParams` in `core/src/types.ts` defines three optional fields beyond the basics:

```typescript
tickSize?: number;   // Minimum price increment override (Limitless/Polymarket)
negRisk?: boolean;   // Skip neg-risk lookup (Polymarket)
onBehalfOf?: number; // Limitless delegated signing: profile ID
```

The **Python SDK** forwards all three correctly. The **TypeScript SDK** builds `paramsDict` in both `createOrder()` and `buildOrder()` and only includes `marketId`, `outcomeId`, `side`, `type`, `amount`, `price`, `fee` — silently dropping `tickSize`, `negRisk`, and `onBehalfOf` before the POST to the sidecar.

## Evidence

`sdks/typescript/pmxt/client.ts`, `createOrder()` (around line 2614):
```typescript
const paramsDict: any = {
    marketId,
    outcomeId,
    side: params.side,
    type: params.type,
    amount: params.amount,
};
if (params.price !== undefined) {
    paramsDict.price = params.price;
}
if (params.fee !== undefined) {
    paramsDict.fee = params.fee;
}
// tickSize, negRisk, onBehalfOf never added
```

Same pattern repeated in `buildOrder()` (around line 2312).

```
$ grep -n "tickSize\|negRisk\|onBehalfOf" sdks/typescript/pmxt/client.ts
(no matches)
```

Python SDK `sdks/python/pmxt/client.py` correctly passes all three:
```python
if tick_size is not None:
    params_dict["tickSize"] = tick_size   # line 3069
if neg_risk is not None:
    params_dict["negRisk"] = neg_risk     # line 3071
if on_behalf_of is not None:
    params_dict["onBehalfOf"] = on_behalf_of  # line 3073
```

## Impact

- **Polymarket**: `negRisk` override always ignored — the exchange must do an extra lookup every time even when the caller knows it's not a neg-risk market.
- **Limitless**: `onBehalfOf` is silently swallowed — delegated profile trading via TypeScript SDK is impossible regardless of what the caller passes.
- **Limitless/Polymarket**: `tickSize` override silently ignored — tick-size resolution falls back to the sidecar's default lookup on every order.

## Expected fix

In both `createOrder()` and `buildOrder()`, after building `paramsDict`, add:
```typescript
if (params.tickSize !== undefined) paramsDict.tickSize = params.tickSize;
if (params.negRisk !== undefined) paramsDict.negRisk = params.negRisk;
if (params.onBehalfOf !== undefined) paramsDict.onBehalfOf = params.onBehalfOf;
```

DeepCogNeural:
## Summary
- Forward `tickSize`, `negRisk`, and `onBehalfOf` in TypeScript SDK local sidecar `buildOrder()` requests.
- Forward the same optional fields in local sidecar `createOrder()` requests.
- Add focused Jest coverage proving both request bodies retain those fields. The test fixes the base URL to `LOCAL_URL` and covers `negRisk: false` plus `onBehalfOf: 0` so falsy values are not dropped.

Fixes #1031

## Validation
- `git diff --check` -> passed.
- `npx jest --runTestsByPath tests/sidecar-order-param-forwarding.test.ts --globals '{"ts-jest":{"diagnostics":false}}'` -> 2 passed.
- `npm test -- --runTestsByPath tests/sidecar-order-param-forwarding.test.ts --runInBand` is blocked before execution by an existing `upstream/main` TypeScript diagnostic: missing `_hostedSubmitOrder` on `Exchange`.
- Existing `tests/hosted-dispatch.test.ts` is blocked by the same diagnostic, confirming this is not introduced by this test.
- Java 17 `npm run build --workspace=pmxtjs` was attempted and is blocked by existing TypeScript errors: missing `_hostedSubmitOrder` and `limitless_buy` schema key.

## Review
- Independent adversarial review found one test-stability issue; addressed by forcing `baseUrl: LOCAL_URL` and adding `onBehalfOf: 0` coverage.
- Follow-up adversarial review had no P0/P2 findings; the remaining P1 is the unrelated upstream TypeScript diagnostic noted above.

### DIFF
--- a/sdks/typescript/pmxt/client.ts
+++ b/sdks/typescript/pmxt/client.ts
@@ -2104,6 +2104,15 @@ export abstract class Exchange {
             if (params.fee !== undefined) {
                 paramsDict.fee = params.fee;
             }
+            if (params.tickSize !== undefined) {
+                paramsDict.tickSize = params.tickSize;
+            }
+            if (params.negRisk !== undefined) {
+                paramsDict.negRisk = params.negRisk;
+            }
+            if (params.onBehalfOf !== undefined) {
+                paramsDict.onBehalfOf = params.onBehalfOf;
+            }
 
             const response = await this.fetchWithRetry(`${this.resolveBaseUrl()}/api/${this.exchangeName}/buildOrder`, {
                 method: 'POST',
@@ -2405,6 +2414,15 @@ export abstract class Exchange {
             }
             if (params.fee !== undefined) {
                 paramsDict.fee = params.fee;
+            }
+            if (params.tickSize !== undefined) {
+                paramsDict.tickSize = params.tickSize;
+            }
+            if (params.negRisk !== undefined) {
+                paramsDict.negRisk = params.negRisk;
+            }
+            if (params.onBehalfOf !== undefined) {
+                paramsDict.onBehalfOf = params.onBehalfOf;
             }
 
             const response = await this.fetchWithRetry(`${this.resolveBaseUrl()}/api/${this.exchangeName}/createOrder`, {
--- a/sdks/typescript/tests/sidecar-order-param-forwarding.test.ts
+++ b/sdks/typescript/tests/sidecar-order-param-forwarding.test.ts
@@ -1,0 +1,104 @@
+import { Polymarket } from "../pmxt/client";
+import { LOCAL_URL } from "../pmxt/constants";
+
+interface CapturedFetch {
+  url: string;
+  init?: RequestInit;
+}
+
+function installFetchSpy(handler: (req: CapturedFetch) => Response): jest.SpyInstance {
+  const captured: CapturedFetch[] = [];
+  const spy = jest.spyOn(global, "fetch").mockImplementation(async (input, init) => {
+    const url = typeof input === "string" ? input : (input as URL | Request).toString();
+    const rec: CapturedFetch = { url, init };
+    captured.push(rec);
+    return handler(rec);
+  });
+  (spy as unknown as { captured: CapturedFetch[] }).captured = captured;
+  return spy;
+}
+
+function captured(spy: jest.SpyInstance): CapturedFetch[] {
+  return (spy as unknown as { captured: CapturedFetch[] }).captured;
+}
+
+function jsonResponse(payload: unknown, status = 200): Response {
+  return new Response(JSON.stringify(payload), {
+    status,
+    headers: { "Content-Type": "application/json" },
+  });
+}
+
+function makePolymarket(): Polymarket {
+  return new Polymarket({ autoStartServer: false, baseUrl: LOCAL_URL });
+}
+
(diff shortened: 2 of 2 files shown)


realfishsam:

### What This Does
Forwards the TypeScript SDK's optional order-building/order-creation params (`tickSize`, `negRisk`, `onBehalfOf`) into the sidecar request body. This matters to TS SDK consumers because these params already exist in the public `CreateOrderParams` shape but were silently dropped before reaching `/api/{exchange}/buildOrder` and `/api/{exchange}/createOrder`.

### Blast Radius
TypeScript SDK `Exchange.buildOrder()` and `Exchange.createOrder()` only, plus a TS Jest regression test that spies on `fetch`. Core sidecar, OpenAPI, Python SDK, and venue implementations are unchanged.

### Consumer Verification
**Before (base branch):**
Static consumer path: the TS SDK constructed `paramsDict` with fields like `marketId`, `outcomeId`, `side`, `type`, `amount`, `price`, `slippagePct`, and `fee`, but did not copy `tickSize`, `negRisk`, or `onBehalfOf`. A consumer calling `new Polymarket(...).createOrder({ ..., tickSize: 0.01, negRisk: false, onBehalfOf: 0 })` would send a sidecar body whose `args[0]` omitted those three keys.

**After (PR branch):**
`buildOrder()` and `createOrder()` now copy all three optional keys when they are not `undefined`; notably `false` and `0` are preserved because the guard is `!== undefined`.

### Test Results
- Build: NOT VERIFIED — TypeScript SDK build is blocked in this checkout because `sdks/typescript/generated/` is absent and OpenAPI generation requires `java`, which is not installed in the review environment.
- Unit tests: NOT VERIFIED — the new `sidecar-order-param-forwarding.test.ts` imports `pmxt/client.ts`, which fails at setup on missing `../generated/src/index.js` before the regression assertions run.
- Server starts: N/A (SDK request-serialization change)
- E2E smoke: NOT VERIFIED — no live order-building endpoint/credentials used

### Findings
No blocking findings.

### PMXT Pipeline Check
- Field propagation (3-layer): N/A — existing fields are forwarded by the SDK shim; no unified response fields added
- OpenAPI sync: N/A
- Financial precision: OK — no arithmetic introduced; values are forwarded as supplied
- Type safety: OK — uses existing typed params
- Auth safety: OK

### Semver Impact
patch -- bugfix for dropped optional SDK request params.

### Risk
Consumer verification is static only in this environment. CI should run the new fetch-spy test after generated TS client artifacts are available; that test is the right consumer-path check for this change.

realfishsam:
Merged! Thanks for contributing!

PR #1308, opened 2026-06-25
fix(sdk): compute detailed execution prices locally

### LINKED ISSUE #1306: `getExecutionPriceDetailed` is a synchronous pure function in core but async with sidecar round-trip in both SDKs
## Summary

`getExecutionPriceDetailed` in `core/src/BaseExchange.ts` is a **synchronous pure function** — it takes order parameters and returns a detailed price breakdown with zero I/O. Both SDKs have reimplemented it as an **async method that makes an HTTP round-trip to the sidecar**, introducing unnecessary latency and a network failure point for what should be a local computation.

## Core implementation (synchronous)

**`core/src/BaseExchange.ts:1202`**

```ts
getExecutionPriceDetailed(
    side: 'buy' | 'sell',
    amount: number,
    orderBook: OrderBook,
    options?: { feeRate?: number; slippage?: number },
): ExecutionPriceResult {
    // Pure in-memory calculation — no I/O
    ...
}
```

Compare with `getExecutionPrice` at line 1190, which is also sync, and which **both SDKs correctly implement as sync**.

## TypeScript SDK (async, calls sidecar)

**`sdks/typescript/pmxt/client.ts:2721`**

```ts
async getExecutionPriceDetailed(
    side: 'buy' | 'sell',
    amount: number,
    orderBook: OrderBook,
    options?: { feeRate?: number; slippage?: number },
): Promise<ExecutionPriceResult> {
    // Makes HTTP POST to /api/:exchange/getExecutionPriceDetailed
    const json = await this.sidecarReadRequest('getExecutionPriceDetailed', ...);
    ...
}
```

The TypeScript SDK is the same language/runtime as the core, so this round-trip is architecturally unnecessary — the same computation could run locally without ever touching the network.

## Python SDK (async, calls sidecar)

**`sdks/python/pmxt/client.py`** — `get_execution_price_detailed` calls `self._call_method("getExecutionPriceDetailed", ...)`, which issues an HTTP request to the sidecar.

Python cannot import TypeScript directly, so a sidecar call is understandable. However it still carries the same observable problems (see below).

## Impact

| | Core | TypeScript SDK | Python SDK |
|---|---|---|---|
| Signature | Synchronous | `async` | Synchronous wrapper over async HTTP |
| Network required | No | **Yes** | **Yes** |
| Can throw on valid inputs | No | **Yes (network error)** | **Yes (network error)** |
| Latency | ~0 ms | 10–100 ms | 10–100 ms |

Concrete problems for SDK users:

1. **Latency**: Every call to `getExecutionPriceDetailed` pays a network round-trip even though no external state is needed — all inputs are passed by the caller.
2. **Surprising failures**: A method documented as a "pure price calculation" can throw `NetworkError` / `ExchangeNotAvailable` even when both inputs are valid and the user has no interest in accessing the exchange.
3. **Signature mismatch**: Core's method is sync; TypeScript SDK's method is `async`. Any code ported between the core and the SDK must be adjusted.
4. **TypeScript-specific**: `getExecutionPrice` (the simpler sibling at `core/src/BaseExchange.ts:1190`) **is** correctly kept sync in the TypeScript SDK. The inconsistency within the same SDK is confusing.

## Expected behaviour

Both SDKs should compute `getExecutionPriceDetailed` locally using the same algorithm as core — no sidecar call needed. For the TypeScript SDK this means porting the calculation client-side (it is pure math on the `OrderBook` type already available locally). For Python this is higher effort but the result is the same: the method should not require a running sidecar or network access to compute a price estimate.

## Files

- `core/src/BaseExchange.ts:1202` — sync reference implementation
- `core/src/BaseExchange.ts:1190` — `getExecutionPrice`, correctly sync in both SDKs
- `sdks/typescript/pmxt/client.ts:2721` — async sidecar call
- `sdks/python/pmxt/client.py` — `get_execution_price_detailed`, sidecar call

realfishsam:
## Summary
- Make TypeScript `getExecutionPriceDetailed` a synchronous local order-book walk instead of a sidecar HTTP call.
- Make Python `get_execution_price_detailed` compute the same pure result locally.
- Add focused TS/Python regressions for full and partial fills with sidecar autostart disabled.

Fixes #1306

## Test Plan
- `git diff --check HEAD~1..HEAD` ✅
- `node sdks/typescript/scripts/generate-client-methods.js` ✅
- `node sdks/python/scripts/generate-client-methods.js` ✅
- `npm test --workspace=pmxtjs -- --runTestsByPath tests/execution-price-detailed-local.test.ts --runInBand` blocked locally: this checkout lacks `sdks/typescript/generated/src/index.js`; `npm run generate --workspace=pmxtjs` is blocked by missing `java` in the cron environment.
- `PYTHONPATH=sdks/python python3 sdks/python/tests/test_execution_price_detailed_local.py` blocked locally: this checkout lacks generated `pmxt_internal`.

### DIFF
--- a/sdks/python/pmxt/client.py
+++ b/sdks/python/pmxt/client.py
@@ -3397,38 +3397,31 @@ class Exchange(ABC):
         Returns:
             Detailed execution result
         """
-        try:
-            # Convert order_book to dict for API call
-            bids = [{"price": b.price, "size": b.size} for b in order_book.bids]
-            asks = [{"price": a.price, "size": a.size} for a in order_book.asks]
-            ob_dict = {"bids": bids, "asks": asks, "timestamp": order_book.timestamp}
+        if amount <= 0:
+            raise ValueError("Amount must be greater than 0")
 
-            body = {
-                "args": [ob_dict, side, amount]
-            }
+        levels = [level for level in (order_book.asks if side == "buy" else order_book.bids) if level.size > 0]
+        levels.sort(key=lambda level: level.price, reverse=(side == "sell"))
 
-            creds = self._get_credentials_dict()
-            if creds:
-                body["credentials"] = creds
+        if not levels:
+            return ExecutionPriceResult(price=0, filled_amount=0, fully_filled=False)
 
-            url = f"{self._resolve_sidecar_host()}/api/{self.exchange_name}/getExecutionPriceDetailed"
+        remaining_amount = amount
+        total_cost = 0.0
+        filled_amount = 0.0
+        epsilon = 0.00000001
 
-            headers = {"Content-Type": "application/json", "Accept": "application/json"}
-            headers.update(self._get_auth_headers())
+        for level in levels:
+            if remaining_amount <= epsilon:
+                break
 
-            response = self._fetch_with_retry(
-                lambda: self._api_client.call_api(
-                    method="POST",
-                    url=url,
-                    body=body,
-                    header_params=headers
-                )
-            )
+            fill_size = min(remaining_amount, level.size)
+            total_cost += fill_size * level.price
+            filled_amount += fill_size
+            remaining_amount -= fill_size
 
-            response.read()
-            data_json = json.loads(response.data)
-
-            data = self._handle_response(data_json)
-            return _convert_execution_result(data)
-        except Exception as e:
-            raise self._parse_api_exception(e) from None
+        return ExecutionPriceResult(
+            price=total_cost / filled_amount if filled_amount > epsilon else 0,
+            filled_amount=filled_amount,
+            fully_filled=remaining_amount <= epsilon,
+        )
--- a/sdks/python/tests/test_execution_price_detailed_local.py
+++ b/sdks/python/tests/test_execution_price_detailed_local.py
@@ -1,0 +1,31 @@
+from pmxt import Polymarket
+from pmxt.models import OrderBook, OrderLevel
+
+
+def test_get_execution_price_detailed_computes_locally_without_sidecar():
+    client = Polymarket(auto_start_server=False)
+    result = client.get_execution_price_detailed(
+        OrderBook(
+            bids=[OrderLevel(price=0.41, size=5), OrderLevel(price=0.43, size=10)],
+            asks=[OrderLevel(price=0.52, size=4), OrderLevel(price=0.5, size=6)],
+        ),
+        "buy",
+        8,
+    )
+
+    assert abs(result.price - 0.505) < 1e-12
+    assert result.filled_amount == 8
+    assert result.fully_filled is True
+
+
+def test_get_execution_price_detailed_reports_partial_fill():
+    client = Polymarket(auto_start_server=False)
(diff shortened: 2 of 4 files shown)


realfishsam:

### What This Does
Moves `get_execution_price_detailed` calculation into the Python and TypeScript SDK clients so consumers can compute detailed fill price from a supplied `OrderBook` without calling the sidecar-only `getExecutionPriceDetailed` route.

### Blast Radius
Python SDK `Exchange.get_execution_price_detailed`, TypeScript SDK `Exchange.getExecutionPriceDetailed`, and their local unit-test coverage. No core exchange normalizer/OpenAPI schema changes.

### Consumer Verification
**Before (base branch):**
The SDK method builds a sidecar request to `POST /api/{exchange}/getExecutionPriceDetailed` with the order book, side, and amount; consumer behavior depends on a running sidecar route.

```python
Polymarket(auto_start_server=False).get_execution_price_detailed(order_book, "buy", 8)
# base implementation attempts sidecar HTTP POST /api/polymarket/getExecutionPriceDetailed
```

**After (PR branch):**
Static review of `sdks/python/pmxt/client.py` and `sdks/typescript/pmxt/client.ts` shows the method now walks asks for buys and bids for sells, sorts the book side by best execution price, returns partial-fill metadata, and raises on non-positive amount. Example from the added tests:

```python
asks=[(0.52, 4), (0.50, 6)], side="buy", amount=8
# fills 6 @ 0.50 + 2 @ 0.52 => price 0.505, filled_amount=8, fully_filled=True
```

Runtime SDK verification was blocked by missing generated SDK artifacts in this checkout, not by this PR's changed logic.

### Test Results
- Unit tests: NOT VERIFIED for SDK tests. `npx jest -c sdks/typescript/jest.config.cjs sdks/typescript/tests/execution-price-detailed-local.test.ts --runInBand` fails before tests run with `TS2307: Cannot find module '../generated/src/index.js'`; `uv run pytest tests/test_execution_price_detailed_local.py -q` fails dependency resolution because project `requires-python >=3.8` conflicts with `urllib3>=2.7.0` requiring Python >=3.10 for the Python 3.9 split.
- Server starts: NOT RUN (SDK-local behavior change; core build passed)
- E2E smoke: NOT VERIFIED (not sidecar-observable after the change)

### Findings
No blocking findings.

### PMXT Pipeline Check
- OpenAPI sync: N/A
- Financial precision: OK for existing SDK numeric model; no new float-sensitive venue/API serialization path added.
- Type safety: OK in changed hand-written code; SDK generated-import blocker prevented full TS test execution.
- Auth safety: N/A

### Semver Impact
patch -- SDK-local bug fix/removal of unnecessary sidecar dependency for an existing helper method.

### Risk
The old TypeScript `getExecutionPrice` helper still walks order-book levels in caller-supplied order, while the new detailed method sorts best-first. That is outside this PR's direct change but means the two helpers can disagree if consumers pass an unsorted book. Full SDK test execution remains unverified until generated artifacts/dependency resolution are available.

Issue #1366, opened 2026-06-28
Response drift: smarkets — events[].end_date is always null; resolution dates fall back to start_datetime


realfishsam:
## Exchange
smarkets

## Severity
LOW

## What Our Normalizer Expects
`core/src/exchanges/smarkets/normalizer.ts:121-125` reads `event.end_date` as the primary source of the market resolution date, falling back to `event.start_datetime` and then the current timestamp:

```typescript
// normalizer.ts:121-125
const resolutionDate = event.end_date
    ? new Date(event.end_date)
    : event.start_datetime
        ? new Date(event.start_datetime)
        : new Date();
```

`end_date` is listed in `SMARKETS_PROMOTED_EVENT_KEYS` (normalizer.ts:23), indicating it is expected to be a meaningful field.

## What The Live API Returns
The live Smarkets events endpoint returns `end_date` as `null` on all observed events:

```
events[].end_date: NoneType   ← null on all sampled events
events[].start_datetime: str  ← present (ISO-8601)
```

Endpoint tested: `GET https://api.smarkets.com/v3/events/?limit=3&state=upcoming`

## Impact
LOW: The fallback prevents crashes — `resolutionDate` is set to `start_datetime` rather than `end_date`. However:
- All Smarkets markets display the **start** time as their resolution date rather than when the market actually closes
- Sorting or filtering by resolution date relative to other venues produces incorrect ordering
- If `start_datetime` is also null, resolution date falls back to the current timestamp at fetch time, which is clearly wrong

## Suggested Fix
Determine whether `end_date` is available on the single-event detail endpoint (`GET /v3/events/{id}/`) when it is absent from the list endpoint. If not available anywhere, update the normalizer to document the limitation and consider using the `start_datetime` + estimated duration as a proxy. Also check whether `modified` or any other event field carries a usable resolution date.

---
*Found by automated response shape drift audit*

PR #1367, opened 2026-06-28
Fix Smarkets resolutionDate fallback

jordansilly77-stack:
## Summary

- stop using `start_datetime` as a Smarkets `resolutionDate` fallback when `end_date` is missing
- stop defaulting missing Smarkets resolution dates to the current time
- add regression coverage for both missing-`end_date` paths

## Context

I checked the Smarkets public events endpoint and the single-event endpoint while looking at #1366. Both can return `end_date: null`; for example, `GET https://api.smarkets.com/v3/events/?limit=3&state=upcoming` and `GET /v3/events/{id}/` returned null `end_date` values for sampled events.

Since `start_datetime` is the event start time rather than a resolution time, exposing it as `resolutionDate` can make cross-venue sorting/filtering look more precise than the venue data supports. `UnifiedMarket.resolutionDate` is optional, so this now leaves it unset unless Smarkets actually provides `end_date`.

## Tests

- `npm test --workspace=pmxt-core -- --runTestsByPath test/normalizers/exchange-normalizers-2.test.ts`
- `npm run build --workspace=pmxt-core`

### DIFF
--- a/core/src/exchanges/smarkets/normalizer.ts
+++ b/core/src/exchanges/smarkets/normalizer.ts
@@ -68,6 +68,14 @@ function buildContractOutcome(
     };
 }
 
+function parseSmarketsResolutionDate(event: SmarketsRawEvent): Date | undefined {
+    // Smarkets list and detail endpoints currently return end_date as null for
+    // many events. start_datetime is an event start, not a market resolution,
+    // so leave the unified resolution date unset instead of using a misleading
+    // fallback that can corrupt cross-venue ordering.
+    return event.end_date ? new Date(event.end_date) : undefined;
+}
+
 // ----------------------------------------------------------------------------
 // Normalizer
 // ----------------------------------------------------------------------------
@@ -117,12 +125,7 @@ export class SmarketsNormalizer implements IExchangeNormalizer<SmarketsRawEventW
             }
         }
 
-        // Derive resolution date from event end_date or start_datetime
-        const resolutionDate = event.end_date
-            ? new Date(event.end_date)
-            : event.start_datetime
-                ? new Date(event.start_datetime)
-                : new Date();
+        const resolutionDate = parseSmarketsResolutionDate(event);
 
         const um = {
             marketId: market.id,
--- a/core/test/normalizers/exchange-normalizers-2.test.ts
+++ b/core/test/normalizers/exchange-normalizers-2.test.ts
@@ -215,6 +215,36 @@ describe('SmarketsNormalizer', () => {
             expect(result.resolutionDate.getFullYear()).toBe(2025);
         });
 
+        it('does not use start_datetime as a fallback resolutionDate when end_date is missing', () => {
+            const rawEventWithoutEndDate: SmarketsRawEvent = {
+                ...rawEvent,
+                end_date: null,
+                start_datetime: '2025-05-01T09:00:00Z',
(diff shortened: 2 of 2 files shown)


realfishsam:
## PR Review: VERIFIED

### What This Does
This changes the Smarkets market normalizer so `resolutionDate` is only populated from the venue `end_date`. If Smarkets omits `end_date`, PMXT now leaves `resolutionDate` unset instead of pretending the event start time (or the current clock time) is the market resolution time.

### Blast Radius
Smarkets market normalization only: `core/src/exchanges/smarkets/normalizer.ts` and the Smarkets fixture coverage in `core/test/normalizers/exchange-normalizers-2.test.ts`. This affects `fetchMarkets` / event enrichment consumers wherever Smarkets markets are returned, including router sorting/filtering paths that rely on unified `resolutionDate`. No SDK shims, OpenAPI schema, auth, or generated files were changed.

### Consumer Verification
**Before (base branch):**
Fixture-equivalent Smarkets raw data with `end_date: null` and `start_datetime: "2025-05-01T09:00:00Z"` normalized to a misleading resolution date:

```bash
node - <<'NODE'
const { SmarketsNormalizer } = require('./core/dist/exchanges/smarkets/normalizer');
// normalizeMarket({ event: { end_date: null, start_datetime: '2025-05-01T09:00:00Z', ... }, ... })
NODE
```

Observed response snippet on `origin/main` after build:

```json
{
  "resolutionDate": "2025-05-01T09:00:00.000Z",
  "title": "UK General Election 2025",
  "marketId": "market-456"
}
```

**After (PR branch):**
The same consumer-path normalizer probe leaves the field absent/undefined when Smarkets has no `end_date`:

```json
{
  "title": "UK General Election 2025",
  "marketId": "market-456"
}
```

The existing `end_date` path is still covered by the normalizer suite (`sets resolutionDate as a Date from event end_date`).

### Test Results
- Unit tests: PASS (`npm test --workspace=pmxt-core`: 797 passed, 3 skipped, 44 suites passed, 1 skipped)
- Server starts: PASS (`GET /health` returned `{"status":"ok",...}`)
- E2E smoke: PASS for the local Smarkets normalization consumer path above; no live Smarkets HTTP smoke was required because the PR changes pure normalization and the fixture matches the fetcher raw type.

### Findings
No blocking findings.

### PMXT Pipeline Check
- Field propagation (3-layer): N/A — no unified type/schema/SDK field was added or renamed; this only changes whether an optional existing field is populated.
- OpenAPI sync: N/A — no BaseExchange/server schema signature changed.
- Financial precision: N/A — no price/amount math changed.
- Type safety: OK — helper returns `Date | undefined`, matching `UnifiedMarket.resolutionDate?: Date`; no new `any` or non-null assertions introduced.
- Auth safety: N/A — no credential or request-signing paths changed.

### Semver Impact
patch -- bug fix to avoid emitting an incorrect optional `resolutionDate` for Smarkets markets that lack a venue resolution/end date.

### Risk
The remaining production risk is venue semantics: this assumes Smarkets `end_date` is the only currently trustworthy resolution/cutoff field among the raw event fields. Markets without `end_date` will now sort/filter as having no resolution date; that is preferable to a fabricated start/current date, but consumers relying on every Smarkets market having a `resolutionDate` may need to handle the already-optional field.

PR #1579, opened 2026-07-09
fix(ts-sdk): export version constant

### LINKED ISSUE #1568: SDK drift: Python exports a `__version__` module constant; TypeScript's `index.ts` has no version export at all
## Drift

Python's package root defines and exposes a version identifier as a plain module attribute; TypeScript's `index.ts` defines no analogous `VERSION`/`version` export (named or on the default `pmxt` object).

## Python SDK

`sdks/python/pmxt/__init__.py:168`:
```python
__version__ = "2.17.1"
```
Accessible as `pmxt.__version__`.

## TypeScript SDK

`sdks/typescript/index.ts` — no equivalent anywhere in the file. The only version string in the TypeScript package lives in `sdks/typescript/package.json` (`"version": "2.18.0"`), which isn't part of the module's exported runtime API.

## Expected

TypeScript should export a version constant (named export and/or on the default export object) so consumers can read it the same way as Python's `pmxt.__version__`.

## Impact

Consumers or diagnostic tooling that want to programmatically read "which SDK version am I using" can do `pmxt.__version__` in Python but have no equivalent in TypeScript without reaching outside the module (e.g. parsing `package.json`). As an incidental note found while verifying this, the two version strings are currently out of sync (`2.17.1` vs `2.18.0`), consistent with them not being derived from a shared source.

---

realfishsam:
## Summary
- Export a TypeScript `__version__` constant from the package root to match the Python SDK public module version surface.
- Keep the value aligned with the Python SDK's current `pmxt.__version__` value.

Fixes #1568

## Test Plan
- `git diff --check`
- `node` + TypeScript `transpileModule` syntax check for `sdks/typescript/index.ts`
- Attempted targeted `tsc` compile; blocked by missing committed generated client artifacts (`sdks/typescript/generated/src/index.js`), so no generated stubs were committed.

### DIFF
--- a/sdks/typescript/index.ts
+++ b/sdks/typescript/index.ts
@@ -26,6 +26,8 @@ import { FeedClient } from "./pmxt/feed-client.js";
 import * as models from "./pmxt/models.js";
 import * as errors from "./pmxt/errors.js";
 
+export const __version__ = "2.17.1";
+
 export { Exchange, Polymarket, Kalshi, KalshiDemo, Limitless, Myriad, Probable, Baozi, Opinion, Metaculus, Smarkets, PolymarketUS, GeminiTitan, Hyperliquid, SuiBets, Suibets, Rain, Hunch, Mock, PolymarketOptions } from "./pmxt/client.js";
 export type { ExchangeOptions, SuiBetsOptions } from "./pmxt/client.js";
 export { FeedClient } from "./pmxt/feed-client.js";


realfishsam:
## PR Review: FAIL

### What This Does
Adds a TypeScript package-root `__version__` export so consumers can inspect the SDK version from `pmxtjs`.

### Blast Radius
TypeScript SDK package root `sdks/typescript/index.ts`; no sidecar/core behavior changes.

### Consumer Verification
**Before (base branch):**
`sdks/typescript/index.ts` did not export `__version__`.

**After (PR branch):**
The PR exports `__version__`, but the value is hardcoded to `"2.17.1"` while `sdks/typescript/package.json` at the PR head is `2.18.0` (`node`/`git show` check in this run). A consumer importing `__version__` from the package root would receive stale version metadata.

### Test Results
- Server starts: N/A for package-root metadata
- E2E smoke: FAIL by package metadata check (`__version__` value does not match package.json)

### Findings
1. `sdks/typescript/index.ts:29` — `export const __version__ = "2.17.1";` is stale relative to `sdks/typescript/package.json` version `2.18.0`. This would immediately publish incorrect SDK version metadata and will drift again on every release unless generated/read from package metadata.

### PMXT Pipeline Check
- OpenAPI sync: N/A
- Financial precision: N/A
- Type safety: OK
- Auth safety: N/A

### Semver Impact
patch -- package metadata export fix, but current value is wrong.

### Risk
Published npm consumers would see the wrong `__version__`, undermining diagnostics and support workflows.

PR #1581, opened 2026-07-09
fix(limitless): throw on watchOrderBook failure instead of silent swallow

### LINKED ISSUE #676: Unhandled async: limitless/websocket.ts — watchOrderBook outer catch swallows all errors silently
## Risk Level
CRITICAL

## Location
`core/src/exchanges/limitless/websocket.ts:195-198`

## Code
```typescript
        } catch (err) {
            // Fallback to empty orderbook if all else fails
            return this.getEmptyOrderbook();
        }
```

This outer `try/catch` wraps the entire body of `watchOrderBook` (lines ~155–198), including the WebSocket connection, the SDK `subscribe` call, the `Promise.race` between a WS update and the REST-snapshot timeout, and all error paths within them.

## What Happens on Failure
Any error thrown anywhere in `watchOrderBook` — WebSocket connection failure, Limitless SDK disconnect, network timeout, unhandled SDK exception — is silently discarded and an empty `{ bids: [], asks: [], timestamp }` object is returned to the caller. **`err` is never logged.**

The streaming loop in `ws-handler.ts` calls `watchOrderBook` in a tight `while` loop. On every iteration it receives an empty orderbook, sends it downstream to WebSocket subscribers, and immediately loops again. The result is:
- Order-routing logic sees zero liquidity on Limitless — mispricing or dropped orders.
- The `streamSingle` loop burns CPU on empty data.
- Operators have zero visibility: no log line, no error frame to the subscriber, no metric.

## Suggested Fix
Log the error before returning the empty fallback, and consider whether an empty orderbook is the right behavior at all (versus letting the error propagate so the streaming loop can break and surface an error frame to the subscriber):

```typescript
} catch (err) {
    logger.error('LimitlessWS: watchOrderBook failed', { error: String(err) });
    // Consider: throw err; so the streaming loop surfaces an error frame
    return this.getEmptyOrderbook();
}
```

---

### LINKED ISSUE #1374: empty-catch: core/src/exchanges/limitless/websocket.ts:195 — `catch (err)` discards error silently in WebSocket orderbook resolver
## Violation

```typescript
        } catch (err) {
            // Fallback to empty orderbook if all else fails
            return this.getEmptyOrderbook();
        }
```

## Location

`core/src/exchanges/limitless/websocket.ts:195–198`

Full context (lines 155–198):
```typescript
try {
    // ... set up WS resolver and timeout race ...
    return await Promise.race([wsUpdatePromise, timeoutPromise]);
} catch (err) {
    // Fallback to empty orderbook if all else fails
    return this.getEmptyOrderbook();
}
```

## Why It Matters

`err` is declared in the catch clause but never referenced — its type, message, and stack trace are silently discarded. The comment documents the *strategy* (return empty orderbook) but not the *cause* (what error triggered the fallback). In production, any unexpected exception thrown during the Promise race setup — a bug in the resolver bookkeeping, an unexpected rejection, or an internal WebSocket state error — will hit this block and produce a silent empty orderbook with zero diagnostic trail. Correlating "why is the Limitless orderbook always empty?" to this catch block requires reading source code rather than logs.

## Suggested Fix

Log the suppressed error at `warn` level before returning the fallback:

```typescript
        } catch (err) {
            logger.warn('LimitlessWS: orderbook watch failed; returning empty', {
                marketSlug,
            });
            return this.getEmptyOrderbook();
        }
```

---

AbhilashG12:
## Description
Fixes #676 and closes #1374.

The `watchOrderBook` method for Limitless was previously catching all SDK/network errors and silently returning an empty orderbook. This caused two severe issues:
1. Upstream stream loops (`streamSingle`) would spin infinitely at max CPU.
2. Downstream trading logic received false "zero liquidity" signals instead of feed degradation alerts.

This PR removes the empty fallback. It now properly logs the error (including the market context as requested in #1374) and throws it, forcing the caller to handle the failure and allowing circuit breakers / reconnect logic to function normally.

### DIFF
--- a/core/src/exchanges/limitless/websocket.ts
+++ b/core/src/exchanges/limitless/websocket.ts
@@ -193,8 +193,13 @@ export class LimitlessWebSocket {
 
             return await Promise.race([wsUpdatePromise, timeoutPromise]);
         } catch (err) {
-            // Fallback to empty orderbook if all else fails
-            return this.getEmptyOrderbook();
+            logger.error('LimitlessWS: watchOrderBook failed', { 
+                market: marketSlug, 
+                error: err instanceof Error ? err.message : String(err)
+            });
+            
+            // Do not return fake liquidity. Force the upstream loop to handle the break.
+            throw err; 
         }
     }
 
--- a/core/src/exchanges/limitless/websocket.ts
+++ b/core/src/exchanges/limitless/websocket.ts
@@ -193,13 +193,13 @@ export class LimitlessWebSocket {
 
             return await Promise.race([wsUpdatePromise, timeoutPromise]);
         } catch (err) {
-            logger.error('LimitlessWS: watchOrderBook failed', { 
-                market: marketSlug, 
+            logger.error('LimitlessWS: watchOrderBook failed', {
+                market: marketSlug,
                 error: err instanceof Error ? err.message : String(err)
             });
-            
+
             // Do not return fake liquidity. Force the upstream loop to handle the break.
-            throw err; 
+            throw err;
         }
     }
 


realfishsam:
## PR Review: FAIL

### What This Does
Attempts to make Limitless `watchOrderBook` throw instead of returning fake empty liquidity when the WebSocket/snapshot path fails. That matters because SDK consumers should not mistake transport failure for an empty order book.

### Blast Radius
Limitless WebSocket order-book path in `core/src/exchanges/limitless/websocket.ts`; no schema or SDK wrapper signature changes.

### Consumer Verification
**Before (base branch):**
Static consumer-path trace: when the timeout fallback `fetchOrderBookSnapshot()` failed, `watchOrderBook()` logged a warning and resolved `getEmptyOrderbook()`, so a consumer could receive empty bids/asks for a transport/API failure.

**After (PR branch):**
The outer `catch` now throws, but the timeout fallback still catches its own snapshot error and resolves `this.getEmptyOrderbook()` at `core/src/exchanges/limitless/websocket.ts:186-189`. That means the concrete failure path in the PR title still returns fake empty liquidity instead of throwing.

### Test Results
- Server starts: NOT VERIFIED for this PR head
- E2E smoke: FAIL by static consumer-path trace for timeout fallback failure

### Findings
1. `core/src/exchanges/limitless/websocket.ts:186-189` — the timeout fallback still swallows `fetchOrderBookSnapshot()` failures and resolves `getEmptyOrderbook()`. The new outer `catch` at lines 195-202 cannot see this because the inner promise resolves successfully, so `watchOrderBook()` still reports an empty order book for this failure mode.

### PMXT Pipeline Check
- OpenAPI sync: N/A
- Financial precision: OK/N/A
- Type safety: OK
- Auth safety: N/A

### Semver Impact
patch -- bug fix intent, but incomplete.

### Risk
If merged as-is, sparse Limitless WS updates plus REST snapshot failure can still produce false zero-liquidity order books for SDK consumers.

PR #1959, opened 2026-08-03
fix(api): add CCXT-compatible params bag to loadMarkets

### LINKED ISSUE #148: loadMarkets: missing trailing params={} bag
## Inconsistency

`loadMarkets` does not accept a trailing `params = {}` bag, making it impossible to pass exchange-specific options during market loading.

## Current (PMXT)

`core/src/BaseExchange.ts`:
```typescript
async loadMarkets(reload: boolean = false): Promise<Record<string, UnifiedMarket>>
```

`sdks/python/pmxt/client.py`:
```python
def load_markets(self, reload: bool = False) -> Dict[str, UnifiedMarket]
```

## Expected (CCXT convention)

`/tmp/ccxt/ts/src/base/Exchange.ts`:
```typescript
async loadMarkets(reload: boolean = false, params: object = {}): Promise<Dictionary<Market>>
```

- Second arg `params: object = {}` allows passing exchange-specific options for the initial market fetch
- Also note: CCXT returns `Dictionary<Market>` (`Record<string, Market>`), consistent with CCXT's `Market` type

## Impact

Minor but breaks signature compatibility. Any CCXT wrapper code that calls `loadMarkets(false, { type: 'spot' })` to narrow the market set will have the `params` object silently ignored by PMXT.

adamantmm:
## Summary

Closes #148.

Adds the trailing `params` argument to `loadMarkets` across core and both SDKs, matching CCXT's `loadMarkets(reload, params)` signature. The bag is forwarded to `fetchMarkets`, so exchange-specific options on the initial market load are no longer silently dropped.

## Changes

- `BaseExchange.loadMarkets(reload, params)` forwards `params` to `fetchMarkets`
- Python SDK: `load_markets(reload, params=None)`
- TypeScript SDK: `loadMarkets(reload, params={})`
- OpenAPI request schema updated (`args` now accepts `[boolean, object]`)
- Unit test covering param forwarding

## Test plan

- [x] `baseExchange.core.test.ts` — verifies `loadMarkets(false, { limit: 1 })` reaches `fetchMarketsImpl`
- [ ] Full `npm test` in CI

Made with [Cursor](https://cursor.com)

### DIFF
--- a/core/src/BaseExchange.ts
+++ b/core/src/BaseExchange.ts
@@ -679,15 +679,19 @@ export abstract class PredictionMarketExchange {
      * `Object.values(exchange.markets)` locally.
      *
      * @param reload - Force a fresh fetch from the API even if markets are already loaded
+     * @param params - Optional exchange-specific fetch parameters (CCXT-compatible trailing bag)
      * @returns Dictionary of markets indexed by marketId
      */
-    async loadMarkets(reload: boolean = false): Promise<Record<string, UnifiedMarket>> {
+    async loadMarkets(
+        reload: boolean = false,
+        params: MarketFetchParams = {}
+    ): Promise<Record<string, UnifiedMarket>> {
         if (this.loadedMarkets && !reload) {
             return this.markets;
         }
 
         // Fetch all markets (implementation dependent, usually fetches active markets)
-        const markets = await this.fetchMarkets();
+        const markets = await this.fetchMarkets(params);
 
         // Reset caches
         this.markets = {};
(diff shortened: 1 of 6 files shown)


adamantmm:
Addressed the review notes:

- widened `MarketFetchParams` with an index signature so exchange-specific keys (e.g. `{ type: 'spot' }`) type-check without casts
- switched the unit test to `{ similarityThreshold: 0.8 }` so it only asserts forwarding, not pagination/slicing
- documented the positional `[reload: boolean, params: object]` order (with example) on the OpenAPI `args` schema in both YAML and JSON

PR #2026, opened 2026-08-12
fix(ts-sdk): align execution price calculation

### LINKED ISSUE #1460: SDK drift: TypeScript `getExecutionPrice` doesn't sort/filter order book levels, unlike `getExecutionPriceDetailed` and unlike Python's `get_execution_price`
## Drift
TypeScript has two independently-implemented execution-price methods that can disagree: `getExecutionPrice` walks `orderBook.asks`/`bids` in whatever order they arrived, with no sort and no filtering of zero-size levels, while the sibling `getExecutionPriceDetailed` explicitly filters out zero-size levels and sorts by price before walking. Python has only one source of truth: `get_execution_price` simply delegates to `get_execution_price_detailed`, which does sort/filter — so the two Python methods can never disagree, but the two TypeScript methods can.

## TypeScript SDK
`sdks/typescript/pmxt/client.ts:2704-2716` (`getExecutionPrice`) — `const levels = side === 'buy' ? orderBook.asks : orderBook.bids;` then a plain unsorted, unfiltered loop. Contrast with `getExecutionPriceDetailed` at `client.ts:2726-2762`, which does `.filter((level) => level.size > 0).sort((a, b) => side === 'buy' ? a.price - b.price : b.price - a.price)` first.

## Python SDK
`sdks/python/pmxt/client.py:3372-3390` (`get_execution_price`) — delegates entirely to `get_execution_price_detailed()` (`client.py:3392-3436`, which sorts/filters), via `result = self.get_execution_price_detailed(order_book, side, amount); return result.price if result.fully_filled else 0`.

## Expected
TypeScript's `getExecutionPrice` should delegate to `getExecutionPriceDetailed` (mirroring Python), or at minimum apply the identical sort/filter logic, so both methods on both SDKs agree on VWAP for the same order book.

## Impact
If a server ever returns `orderBook.asks`/`bids` that aren't already best-price-first, or that include zero-size levels, `getExecutionPrice` and `getExecutionPriceDetailed` will return materially different prices for the identical order book in TypeScript — a correctness bug with direct trading/pricing impact. This cannot happen in Python.

---

Purin1410:
## Summary

- Reuse `getExecutionPriceDetailed` in the TypeScript helper.
- Sort levels consistently, ignore non-positive sizes, and preserve the `0` result when liquidity is insufficient.

## Test plan

- `npm test --workspace=pmxtjs -- --runInBand tests/execution-price-detailed-local.test.ts`

## Note

Amounts less than or equal to zero now use the detailed helper's existing validation behavior.

Fixes #1460

### DIFF
--- a/sdks/typescript/pmxt/client.ts
+++ b/sdks/typescript/pmxt/client.ts
@@ -2894,17 +2894,8 @@ export abstract class Exchange {
      * @returns The volume-weighted average price, or 0 if insufficient liquidity
      */
     getExecutionPrice(orderBook: OrderBook, side: 'buy' | 'sell', amount: number): number {
-        const levels = side === 'buy' ? orderBook.asks : orderBook.bids;
-        let remaining = amount;
-        let totalCost = 0;
-        for (const level of levels) {
-            const fill = Math.min(remaining, level.size);
-            totalCost += fill * level.price;
-            remaining -= fill;
-            if (remaining <= 0) break;
-        }
-        if (remaining > 0) return 0;
-        return totalCost / amount;
+        const result = this.getExecutionPriceDetailed(orderBook, side, amount);
+        return result.fullyFilled ? result.price : 0;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


Purin1410:
Replaced by #2029 after renaming the source branch.

PR #2027, opened 2026-08-12
fix(ts-sdk): use shared transport for market comparison

### LINKED ISSUE #1485: SDK drift: `compareMarketPrices` bypasses the shared retrying transport in TypeScript, unlike Python and unlike its own TS sibling Router methods
## Drift
`compareMarketPrices` alone issues a raw `fetch()` call directly instead of going through the shared `sidecarReadRequest`/`fetchWithRetry` transport that every other Router method uses — so it lacks the connection-failure retry logic (3 retries, 200/500/1000ms backoff) its siblings get for free. Python is uniform: `compare_market_prices` uses the exact same `_call_method` helper (with retry) as every other Router method.

## TypeScript SDK
`sdks/typescript/pmxt/router.ts:476-490` — raw `await fetch(url, {...})`, duplicating its own `AbortSignal.timeout(30_000)` and a dynamic `import('./errors.js')` for error parsing, none of which the other Router methods do.

## Python SDK
`sdks/python/pmxt/router.py:477` — `compare_market_prices` calls `self._call_method("compareMarketPrices", params)`, identical to every other Router method.

## Expected
`compareMarketPrices` in TypeScript should route through the same `sidecarReadRequest`/`fetchWithRetry` transport as its sibling Router methods.

## Impact
A transient connection failure that every other Router method (and every Python Router method, including `compare_market_prices`) would transparently retry causes `compareMarketPrices` in TypeScript alone to fail immediately — an inconsistent reliability characteristic for what should be a uniform set of read operations.

---

Purin1410:
## Summary

- Route `Router.compareMarketPrices` through the shared sidecar read transport.
- Add dynamic URL resolution, connection retries, and the existing compatibility fallback without changing response mapping.

## Test plan

- `npm test --workspace=pmxtjs -- --runInBand tests/router-compareMarketPrices.test.ts`

Fixes #1485

### DIFF
--- a/sdks/typescript/pmxt/router.ts
+++ b/sdks/typescript/pmxt/router.ts
@@ -477,32 +477,12 @@ export class Router extends Exchange {
         if (params.url) query.url = params.url;
 
         try {
-            const url = `${this.config.basePath}/api/${this.exchangeName}/compareMarketPrices`;
-            const response = await fetch(url, {
-                method: 'POST',
-                headers: { 'Content-Type': 'application/json', ...this.getAuthHeaders() },
-                body: JSON.stringify({ args: [query], credentials: this.getCredentials() }),
-                signal: AbortSignal.timeout(30_000),
-            });
-            if (!response.ok) {
-                const body = await response.json().catch(() => ({}));
-                if (body.error && typeof body.error === 'object') {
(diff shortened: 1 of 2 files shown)


Purin1410:
Replaced by #2030 after renaming the source branch.

PR #2028, opened 2026-08-12
fix(ts-sdk): exclude null price changes from filters

### LINKED ISSUE #1524: SDK drift: `filterMarkets`/`filter_markets` handle an explicit `null` `priceChange24h` differently — TypeScript's `undefined` check lets it slip through
## Drift

The `priceChange24h`/`price_change_24h` filter's exclusion guard checks for `undefined` in TypeScript but `None` in Python. These are not equivalent: a market outcome whose `priceChange24h` is explicitly `null` in the server JSON deserializes to JS `null` (not `undefined`), so TypeScript's guard does not catch it.

## TypeScript SDK

`sdks/typescript/pmxt/client.ts:2972-2981` (`filterMarkets`):
```typescript
if (criteria.priceChange24h) {
    const outcome = market[criteria.priceChange24h.outcome];
    if (!outcome || outcome.priceChange24h === undefined) return false;

    if (criteria.priceChange24h.min !== undefined && outcome.priceChange24h < criteria.priceChange24h.min) {
        return false;
    }
    ...
}
```

## Python SDK

`sdks/python/pmxt/client.py:2131-2139` (`filter_markets`):
```python
if "price_change_24h" in params:
    f = params["price_change_24h"]
    outcome_key = f.get("outcome")
    if outcome_key:
        outcome = getattr(market, outcome_key, None)
        if not outcome or outcome.price_change_24h is None: continue
        if "min" in f and outcome.price_change_24h < f["min"]: continue
        ...
```

## Expected

Both SDKs should exclude a market from a `priceChange24h`/`price_change_24h` filter when the field is explicitly `null`/`None`, matching Python's behavior (TypeScript's check should be `=== undefined || === null`, or equivalently use `== null`).

## Impact

For a market outcome whose `priceChange24h` is explicitly `null` (not yet calculable), Python's `filter_markets` correctly excludes it. TypeScript's `=== undefined` check does not catch `null`, so execution falls through to `null < criteria.priceChange24h.min`, which JS coerces via `null` → `0`. Depending on the configured `min`/`max`, a market that should have been excluded entirely can incorrectly pass (or fail) the filter — producing different filtered result sets between the two SDKs for identical input data.

---

Purin1410:
## Summary

- Exclude an explicit `null` `priceChange24h` from price-change filters.
- Preserve numeric zero values and add a regression test for the distinction.

## Test plan

- `npm test --workspace=pmxtjs -- --runInBand tests/filter-markets-local.test.ts`

Fixes #1524

### DIFF
--- a/sdks/typescript/pmxt/client.ts
+++ b/sdks/typescript/pmxt/client.ts
@@ -3112,7 +3112,7 @@ export abstract class Exchange {
             // Price change filter
             if (criteria.priceChange24h) {
                 const outcome = market[criteria.priceChange24h.outcome];
-                if (!outcome || outcome.priceChange24h === undefined) return false;
+                if (!outcome || outcome.priceChange24h == null) return false;
 
                 if (criteria.priceChange24h.min !== undefined && outcome.priceChange24h < criteria.priceChange24h.min) {
                     return false;
--- a/sdks/typescript/tests/filter-markets-local.test.ts
+++ b/sdks/typescript/tests/filter-markets-local.test.ts
@@ -1,0 +1,31 @@
+import { Polymarket } from '../pmxt/client';
+
+describe('filterMarkets', () => {
+    it('filters out markets where priceChange24h is null while keeping zero', () => {
+        const client = new Polymarket({ autoStartServer: false });
+
+        const nullPriceChangeMarket: any = {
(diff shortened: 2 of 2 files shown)


Purin1410:
Replaced by #2031 after renaming the source branch.

